PR #23, opened 2015-02-18
Batch for DDNS updated - added support for 2 Step Verification

yonathan9669:
I'll like to contribute

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 output
 tmp
 crash.log
+.idea/
--- a/content/ddns/ddns.sh
+++ b/content/ddns/ddns.sh
@@ -1,14 +1,20 @@
 #!/bin/bash
- 
+
 LOGIN="your@email"
-TOKEN="your-api-token"
+ATOKEN="your-api-token"
+DTOKEN="your-domain-token"
+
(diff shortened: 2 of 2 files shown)


weppos:
Hello @yonathan9669, can you clarify the purpose of this PR? The code is incorrect, you should not use the domain token and API token together.

yonathan9669:
Hello @weppos, i create my dnsimple account 2 months ago and when i try to use the script for refresh a record as a DDNS service i couldn't. I try the differents tools listed on the site without a result, so 3 days ago i decide to check and fix it by my self. I found that the dnsimple api server respond something like authentication problem, thats when i start to check the complete HEAD response. I have the 2 step verification access active, so i just start to make some probes and figure it out that the server need both the API TOKEN and the DOMAIN TOKEN to let me update the record. I make the same changes to the nodejs tool listed (and fix another issues) and make a pull request too, the author didn't answer yet.

weppos:
@yonathan9669 if you use the API token, it will work even if you enable 2FA without changing anything. The changes to 2FA only affects authentication via username and password.

The issue must have been something different. The change as it is right now, it's invalid.

If you are experiencing an issue with the current code, let us know the request and response details and we'll help you to debug the issue.

PR #27, opened 2015-08-11
Point User to Proper API Token Location

nicksteele:
The User's API Token is shown on https://dnsimple.com/user rather than https://dnsimple.com/account


weppos:
Thanks for reporting it! Merged and deployed.

PR #28, opened 2015-09-16
fix overview link 404

3flex:
The Overview link currently links to /v1/overview/ which doesn't exist.
The Overview actually lives at /v1/


weppos:
Thanks!

Issue #31, opened 2015-10-08
Match style of the support site


sbastn:
Bring from the support site:
* A label that identifies the site as "developer".
* Typography: font-stack, styles.
* Colors
* Categories list on the left.
* High-res logo.

PR #32, opened 2015-10-08
Improve typography and style

sbastn:
This PR update the style and typography to match the support site.

Based on #31.

### Before
screen shot 2015-10-08 at 7 47 19 pm

### Now
screen shot 2015-10-08 at 7 45 40 pm


weppos:
@sbastn keep in mind we currently have 2 layouts. It seems in the PR you just updated the default. The other is a clone we temporarily use for v2 API docs.

sbastn:
Oh, didn't know about that one. Updating!

sbastn:
Updated both layouts

PR #34, opened 2015-10-14
Api v2 pagination

aeden:
Documents responses, errors (a start) and pagination.

### DIFF
--- a/content/v2.markdown
+++ b/content/v2.markdown
@@ -63,28 +63,13 @@ All successful responses will be returned as a JSON object with at least one key
 
 The `data` element will contain either a single JSON object or a list of JSON objects, depending on the endpoint.
 
-```
+```json
 {
-    "data": {
-        "id": 1,
-        "account_id": 1,
-        "label": "",
-        "first_name": "Jane",
-        "last_name": "Smith",
-        "job_title": "",
-        "organization_name": "",
-        "email_address": "jane.smith@gmail.com",
-        "phone": "+1 1234567890",
-        "fax": "",
-        "address1": "111 SW 1st Street",
-        "address2": "",
-        "city": "Miami",
-        "state_province": "FL",
-        "postal_code": "33143",
-        "country": "US",
-        "created_at": "2015-05-14T10:05:46.618Z",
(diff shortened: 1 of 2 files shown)


content/v2.markdown
```diff
@@ -57,3 +57,120 @@ curl -H 'X-DNSimple-Domain-Token: DOMAIN-TOKEN' \
...
+```
+
+You may include the following arguments as query parameters when you call an API endpoint that supports pagination:
+
+- `page`: The page to return (default: 1)
+- `per_page`: The number of entries to return per page (default: 30)
+
+For example:
+
+    https://api.dnsimple.com/v2/domains?per_page=100&page=2
```
weppos: This URL is incorrect. It should be

    https://api.dnsimple.com/v2/:account/domains?per_page=100&page=2

I'm not sure if it's better to simply use

    https://api.dnsimple.com/v2/account/domains?per_page=100&page=2

or a random value such as

    https://api.dnsimple.com/v2/1/domains?per_page=100&page=2

content/v2.markdown
```diff
@@ -57,3 +57,120 @@ curl -H 'X-DNSimple-Domain-Token: DOMAIN-TOKEN' \
...
 
+## Responses
+
+All successful responses will be returned as a JSON object with at least one key: `data`.
+
+The `data` element will contain either a single JSON object or a list of JSON objects, depending on the endpoint.
+
+```
+{
+    "data": {
```
weppos: In all the examples I've been using 2 chars indentation. Would you mind to adjust them?

I also suggest here and the `data` example below, to use just a small representation of the object rather a full one so that it will not distract the reader from the real focus.

```
{
  "data": [
    {
      "id": 1,
      "first_name": "Jane",
      "last_name": "Smith",
    },
    {
      "id": 3,
      "first_name": "John",
      "last_name": "Smith",
    }
  ],
  "pagination": {
    "current_page": 1,
    "per_page": 30,
    "total_entries": 2,
    "total_pages": 1
  }
}
```

weppos:
:shipit:

PR #38, opened 2015-12-01
Document the domain name servers API v2 endpoint.

aeden:
Documents the two end points for domain name server management.


weppos:
Remember to add the link to the page in the sidebar. In the v1 this represents a whole new section called `nameservers`. I'd rename it to `name-servers` for now.

As a side note, I'm still not sure how to properly route these URLs, considering we also have the vanity name servers and the endpoint to register the name servers.

I suggest as a next step we plan all these URLs so that they are reasonably consistent now that we have the chance to change them in API v2.

aeden:
@weppos I propose moving the domain name servers into a subdirectory `domain`, and in fact I have updated the PR to reflect this.

weppos:
@aeden I'm fine for now. However, FYI it was under domain in the past, we then decided to promote it as top-level. See 08a8b004cf6c775c3e14c0307c023cebce3581ab

aeden:
Yes, you decided to change it. :trollface:

As you've mentioned before, regardless of whether or not it stays at the top level, we'll need different naming for the other name server-related operations I think.

weppos:
> Yes, you decided to change it.

It was as a result of a discussion in Slack/Hipchat where we all agreed (or nobody disagreed) :smirk:

aeden:
Evidence is required. :stuck_out_tongue_winking_eye:

aeden:
Moving discussion of sidebar structure to https://github.com/aetrion/dnsimple-developer/issues/39

PR #40, opened 2015-12-02
API v2 services and domain services

jodosha:


content/v2/services.markdown
```diff
@@ -0,0 +1,88 @@
...
+    "name": "Google Apps",
+    "short_name": "googleapps",
+    "description": "Add Google Apps for your domain.",
+    "created_at": "2015-11-23T11:40:05Z",
+    "updated_at": "2015-11-23T11:40:05Z",
+  }
+}
+~~~
+
+Responds with HTTP 404 if not found.
```
weppos: The 404 response is implicit in any API method, I'd take it out.

content/v2/services.markdown
```diff
@@ -0,0 +1,88 @@
...
+excerpt: This page documents the DNSimple services API v2.
+---
+
+# Services API
+
+* TOC
+{:toc}
+
+Services provide a way to apply many records to a domain at once.
+Additionally any changes made to a service will automatically be reflected in the domain's records.
```
weppos: This specific line is incorrect, as at the time being we don't propagate the changes when a service is updated. I'd remove it. In fact, we may probably want to remove both lines for now as we haven't really described any of the Product in any API page so far.

weppos:
Other than the changes I commented inline, I'd say :shipit:

PR #44, opened 2015-12-11
Add section about rate limiting

jacegu:
Overview of the important information about rate limiting in API v2

### DIFF
--- a/content/v2.markdown
+++ b/content/v2.markdown
@@ -51,8 +51,7 @@ Note that Oauth2 tokens cam be acquired [programmatically and via OAuth dance](/
 ##### DNSimple Domain Token
 
 ~~~
-curl -H 'X-DNSimple-Domain-Token: DOMAIN-TOKEN' \
-     https://api.dnsimple.com/v2/_/domains/example.com
+curl -H 'X-DNSimple-Domain-Token: DOMAIN-TOKEN' https://api.dnsimple.com/v2/_/domains/example.com
 ~~~
 
 The [Authentication](/v2/authentication/) page provides more information about how to use them.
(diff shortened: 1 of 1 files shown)


content/v2.markdown
```diff
@@ -129,3 +129,50 @@ For example:
...
 
 
+## Rate limiting
+
+There is a limit to the number of requests that you can perform per hour.
+
+- For **authenticated requests** you can make **up to 4,000 requests per hour**.
+- For **unauthenticated requests**, you can make **up to 30 requests per hour**.
+
+Note that both authenticated and unauthenticated requests are associated with your IP address, and not the user or account making them.
```
weppos: @jacegu I'm not sure I'd mention this as, especially for authenticated requests, we'll probably change over a token-based or account-based throttling at some point to avoid abuse or side effects of having client sending us requests from multiple IPs or the multiple clients from the same IP.

I propose to remove this line for now.

content/v2.markdown
```diff
@@ -129,3 +129,48 @@ For example:
...
+
+There is a limit to the number of requests that you can perform per hour.
+
+- For **authenticated requests** you can make **up to 4,000 requests per hour**.
+- For **unauthenticated requests**, you can make **up to 30 requests per hour**.
+
+You can see your current rate limit status by checking the HTTP headers of any API request:
+
+~~~
+$ curl -H "Authorization: Bearer XXXXXXX" -I "https://api.dnsimple.com/v2/whatever"
```
weppos: I'd use `https://api.dnsimple.com/v2/whoami` so that the user can try it.

weppos:
:shipit:

PR #45, opened 2015-12-17
Document the whoami API v2 endpoint.

aeden:


weppos:
It looks good to me. :+1:

PR #46, opened 2015-12-17
Templates API v2 documentation

aeden:


content/v2/templates.markdown
```diff
@@ -0,0 +1,199 @@
...
+## Delete a template {#delete}
+
+    DELETE /:account/templates/:template
+
+### Parameters
+
+Name | Type | Description
+-----|------|------------
+`:account` | `integer` | The account id or wildcard (`_`)
+`:template` | `integer` | The template id or short name
```
weppos: It should probably be

    `:template` | `integer`, `string` | The template id or short name

content/v2/templates.markdown
```diff
@@ -0,0 +1,199 @@
...
+
+### Input
+
+See [create](#create).
+
+### Response
+
+Responds with HTTP 200 on success.
+
+~~~js
```
weppos: I suggest we use

    ~~~json

aeden: Odd, I pulled this directly from the contacts documentation (I think).

weppos: I've just noticed it recently in a previous commit. It may be possible we'll have to update also some other files.

PR #48, opened 2015-12-31
Document OAuth 2 for API v2

weppos:
This PR documents the web application flow.

### DIFF
--- a/content/v2/oauth.markdown
+++ b/content/v2/oauth.markdown
@@ -70,6 +70,17 @@ The following values should be passed as POST parameters:
 
 You'll receive a JSON response. If the request is successful, the response will include an access token, the token type and the account ID. The token type will always be `bearer`.
 
+##### Example Response
+
+~~~json
+{
+  "access_token": "ACCESS_TOKEN",
+  "account_id": 1010,
+  "token_type": "bearer",
+  "scope": null
+}
+~~~
+
 ### Step 3 - API authentication
 
 The access token allows you to execute authenticated API requests on a behalf of the user account. When you'd like to make API calls to DNSimple, simply include the authorization header with each request.
--- a/content/v2/oauth.markdown
+++ b/content/v2/oauth.markdown
@@ -24,6 +24,8 @@ The API currently supports only the OAuth 2 [Authorization Code Grant](http://to
 
 ### Step 1 - Authorization
 
+This starts the OAuth 2.0 authorization flow. This isn't an API call — it's the web page that lets the user sign in to DNSimple and authorize your app. After the user decides whether or not to authorize your app, they will be redirected to the URI specified by `redirect_uri`.
+
 Your web or mobile app should redirect users to the following URL:
 
 ~~~
--- a/content/v2/oauth.markdown
+++ b/content/v2/oauth.markdown
@@ -29,14 +29,14 @@ This starts the OAuth 2.0 authorization flow. This isn't an API call — it's th
 Your web or mobile app should redirect users to the following URL:
 
 ~~~
-https://dnsimple.com/oauth/authorize
+GET https://dnsimple.com/oauth/authorize
 ~~~
 
 The following values should be passed as GET parameters:
 
 `response_type` | **Required**. The grant type requested. We currently only support `code`.
 `client_id`     | **Required**. The client ID you received from DNSimple when you registered the application.
-`redirect_uri`  | Where to redirect the user after authorization has completed. This must be the exact URI registered or a subdirectory. See details below about [redirect urls](#redirect-urls).
+`redirect_uri`  | Where to redirect the user after authorization has completed. This must be the exact URI registered or a subdirectory.
 `scope`         | We currently don't use this field.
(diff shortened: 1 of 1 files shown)


weppos:
The initial documentation is ready. In a next step I'll expand the information about the redirect URI and other validations, but I believe that at this time it contains enough information to start playing with it.

I'll also review our internal implementation to make sure I didn't forget something.

/cc @aetrion/dnsimple for review. In particular, @jacegu and @sbastn let me know if anything is missing given you already had the chance to use it as a client.

content/v2/oauth.markdown
```diff
@@ -16,15 +16,63 @@ Developers need to register their application before getting started. A register
...
+
+~~~
+GET https://dnsimple.com/oauth/authorize
+~~~
+
+The following values should be passed as GET parameters:
+
+`response_type` | **Required**. The grant type requested. We currently only support `code`.
+`client_id`     | **Required**. The client ID you received from DNSimple when you registered the application.
+`redirect_uri`  | Where to redirect the user after authorization has completed. This must be the exact URI registered or a subdirectory.
```
jacegu: I think this needs further clarification:

> This must be the exact URI registered **as callback** or a subdirectory.

I'm also not sure _subdirectory_ is the correct term here but can't think of anything better right now.

content/v2/oauth.markdown
```diff
@@ -16,15 +16,63 @@ Developers need to register their application before getting started. A register
...
+`scope`         | We currently don't use this field.
+`state`         | An unguessable random string. It is used to protect against cross-site request forgery attacks and it will be passed back to your redirect URI.
+
+Because `/oauth/authorize` is a website, there is no direct return value. However, after the user authorizes your app, they will be sent to your redirect URI.
```
jacegu: > they will be sent to your redirect URI.

It is probably better to say:

> they will be redirected to the URI specified in the `redirect_uri` parameter.

content/v2/oauth.markdown
```diff
@@ -16,15 +16,63 @@ Developers need to register their application before getting started. A register
...
+
+If an error occurs, including if the user has chosen not to authorize the app, the following parameters will be included in the redirect URI:
+
+`error`             | An error code per [Section 4.1.2.1 of the OAuth 2.0 spec](http://tools.ietf.org/html/rfc6749#section-4.1.2.1).
+`error_description` | A user-friendly description of the error that occurred.
+`state`             | The state content, if any, originally passed to `/oauth/authorize`.
+
 ### Step 2 - Access Token
 
+This API method is used to exchange the `code` with a bearer token you can use to authenticate to the DNSimple API.
```
jacegu: > to exchange the `code` with a bearer token

We probably need a :us: for this one but I think we should say instead

> to exchange the `code` **for** a bearer token

content/v2/oauth.markdown
```diff
@@ -16,15 +16,63 @@ Developers need to register their application before getting started. A register
...
+~~~
+POST https://dnsimple.com/oauth/access_token
+~~~
+
+The following values should be passed as POST parameters:
+
+`client_id`     | **Required**. The client ID you received from DNSimple when you registered the application.
+`client_secret` | **Required**. The client secret you received from DNSimple when you registered the application.
+`code`          | **Required**. The code acquired in the previous authorization step.
+`redirect_uri`  | Only used to validate that it matches the original `/oauth/authorize`, not used to redirect again.
```
jacegu: No `state` allowed here?

content/v2/oauth.markdown
weppos: I was in the same situation. I'll add a section later that explains all the valid and invalid cases for the `redirect_uri`.

content/v2/oauth.markdown
```diff
@@ -16,15 +16,63 @@ Developers need to register their application before getting started. A register
...
+
+You'll receive a JSON response. If the request is successful, the response will include an access token, the token type and the account ID. The token type will always be `bearer`.
+
```
jacegu: Should we provide an example of the response (even if it is a super simple one)?

content/v2/oauth.markdown
weppos: @jacegu your sentence is a little bit misleading. In theory, you can leave the `redirect_uri` parameter out and we'll use the URI you registered when you registered the app.

jacegu:
@weppos I've suggested a few minor improvements but I think the article covers everything. 

:+1:

content/v2/oauth.markdown
weppos: Good question, I need to check. At this point the state should be useless because you already have a one-time `code` (which guarantees against forged requests).

content/v2/oauth.markdown
weppos: Yes. I'm considering to move this method into the API and document it as an API method.

weppos:
Thanks. I'll quickly review the open questions and I'll add an example.

content/v2/oauth.markdown
weppos: I confirm that `state` is ignored here.

PR #52, opened 2016-01-29
API v2: Allow array as name servers update payload

jodosha:


content/v2/domains/name-servers.markdown
```diff
@@ -74,7 +74,21 @@ Name | Type | Description
 ##### Example
 
 ~~~json
-{}
+"{
+  "name_servers": [
+    "ns1.example.com",
+    "ns2.example.com"
+  ]
+}"
```
weppos: @jodosha I think we should just support one way, unless it proves to be confusing. I'd only document (and support) the array version.

weppos:
@jodosha any proposal on how to update the following section?

```
Name | Type | Description
-----|------|------------
`name_servers` | `array` | **Required** A list of name server names as strings.
```

jodosha:
@weppos Should it be unnamed?

weppos:
I wonder if we should replace the entire section with

```
**Required** A list of name server names as a JSON-encoded array of strings.
```

jodosha:
@weppos That would lack of consistence with other payload descriptions. My suggestion is:

```
```
Name | Type | Description
-----|------|------------
 | `array` | **Required** A list of name server names as strings.
```
```

weppos:
I thought Markdown would complain about a table with an empty content. I'm fine to go that way.

jodosha:
@weppos There is an empty space, btw :wink:

jodosha:
screen shot 2016-01-29 at 15 17 34

This is the rendering.

weppos:
:+1: deployed

PR #57, opened 2016-02-15
API v2: TLDs

jodosha:


content/v2/tlds.markdown
```diff
@@ -0,0 +1,221 @@
...
+
+    curl -H 'Authorization: Bearer <token>' \
+         -H 'Accept: application/json' \
+         https://api.dnsimple.com/v2/tlds/ca/extended_attributes
+
+### Response
+
+~~~json
+{
+  "data"=>[
```
weppos: This is not JSON :smile: 

FYI, just leave the response empty, I've been reusing the fixtures we created for the API client. See 13e7832380f286ded33c636e87c43a3d866006a7

jodosha: Ouch, forgot to replace there. :sweat_smile: Ok, I'll leave these empty.

weppos:
LGTM :+1:

weppos:
Merged but not deployed, I'll wait for the feature to be merged in the app. :)

PR #63, opened 2016-03-24
Clarify sandbox vs production.

aeden:
Hopefully will make it easier to know when to use each (sandbox for testing, production for live data).


weppos:
@aeden I've just made a similar change, and pushed a branch.

**Before**

screen shot 2016-03-24 at 10 53 33

**After**

screen shot 2016-03-24 at 10 53 25

Should I merge the two changes?

aeden:
Just use your branch, it covers everything.

PR #66, opened 2016-05-02
API v2: document filters for domains, zones, and zone records

jodosha:


content/v2/domains.markdown
```diff
@@ -37,6 +37,20 @@ Responds with HTTP 200.
 <%= pretty_print_fixture("/listDomains/success.http") %>
 ~~~
 
+### Filters
+
+Name | Type | Description
+-----|------|------------
+`:name_like` | `wildcard` | Wildcard search for domain name
```
weppos: I'd slightly change the description to be more understandable to whoever was not involved in the decision process. Something like:

> Only include domains containing given string.

I would also remove the "type" concept, as this is something somehow internal.

jodosha: @weppos Please have a look at https://github.com/aetrion/dnsimple-developer/pull/66/commits/234a3956c4f905cc673527698703cf159e09d839

weppos:
@jodosha it looks better to me

Issue #61, opened 2016-03-14
Add example responses where they are missing


aeden:
- [x] https://developer.dnsimple.com/v2/zones/records/ (#68)
- [x] https://developer.dnsimple.com/v2/services/ (#65)
- [x] https://developer.dnsimple.com/v2/templates/  (bd16a14327e844917334ef5b504d5e60c5898873 and 0996fb4db2e41059137733d4161662dc197053e1)
- [x] https://developer.dnsimple.com/v2/templates/records/ (#67)
- [x] https://developer.dnsimple.com/v2/vanity/

jacegu:
I have updated the description of the issue to track progress

weppos:
Completed.

PR #68, opened 2016-05-03
Include zone record fixtures in docs

jacegu:
Related to #61 

The fixtures where there but they had not been added to the docs.


jacegu:
@sbastn says :shipit: 

jazzhands

PR #70, opened 2016-05-05
fix typo in delete zone URL

3flex:


jodosha:
@3flex Hello, thanks for this PR. Merging right away! 👍

PR #72, opened 2016-05-10
API v2: Document sorting (part 1)

jodosha:
/cc @weppos @jacegu

### DIFF
--- a/content/v2.markdown
+++ b/content/v2.markdown
@@ -173,7 +173,7 @@ List all domains for the account 1010 sorting them by expiration date (asc) and 
 
     curl  -H 'Authorization: Bearer <token>' \
           -H 'Accept: application/json' \
-          https://api.dnsimple.com/v2/1010/domains?sorting=expires_on:asc,id:desc
+          https://api.dnsimple.com/v2/1010/domains?sort=expires_on:asc,id:desc
 
 ## Rate limiting
 
--- a/content/v2.markdown
+++ b/content/v2.markdown
@@ -159,12 +159,13 @@ For example:
 
 ## Sorting
 
-API v2 results are implictely sorted according to policies that vary from endpoint to endpoint.
+API v2 results are implicitly sorted according to policies that vary from endpoint to endpoint.
 
 You can decide your own sorting policy for each single API call via the `sort` param.
 This param accepts a pair of informations: the name of a field and the order (`asc` for ascending and `desc` for descending).
 
-Multiple fields are allowed and MUST be comma separated.
+This param accepts a set of comma separated key-value pairs: the name of a field and the order criteria (`asc` for ascending and `desc` for descending).
+
 The order of fields is relevant, as it will determine the priority of the sorting policies.
 
 ### Example
--- a/content/v2.markdown
+++ b/content/v2.markdown
@@ -161,10 +161,10 @@ For example:
 
 API v2 results are implicitly sorted according to policies that vary from endpoint to endpoint.
 
-You can decide your own sorting policy for each single API call via the `sort` param.
-This param accepts a pair of informations: the name of a field and the order (`asc` for ascending and `desc` for descending).
+You can decide your own sorting policy for each single API call via the `sort` parameter.
+This parameter accepts a pair of informations: the name of a field and the order (`asc` for ascending and `desc` for descending).
 
-This param accepts a set of comma separated key-value pairs: the name of a field and the order criteria (`asc` for ascending and `desc` for descending).
+This parameter accepts a set of comma separated key-value pairs: the name of a field and the order criteria (`asc` for ascending and `desc` for descending).
 
 The order of fields is relevant, as it will determine the priority of the sorting policies.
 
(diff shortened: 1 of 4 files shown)


content/v2/domains.markdown
```diff
@@ -52,6 +52,29 @@ List all domains for the account 1010 that have name matching "example":
           -H 'Accept: application/json' \
           https://api.dnsimple.com/v2/1010/domains?name_like=example
 
+### Sorting
+
+Sort records with the `sorting` param by specifing the field and the sorting policy: `asc` for ascending and `desc` for descending.
+Multiple fields are allowed and MUST be comma separated.
```
weppos: Should we change

> Multiple fields are allowed and MUST be comma separated.

into

> Multiple fields are allowed and MUST be comma separated. The order of fields is relevant, as it will determine how the `ORDER` SQL clause is constructed.

weppos: Maybe even something less technical, but something that explains that the sorting will be based on the order of the fields in the parameter.

content/v2/domains.markdown
```diff
@@ -52,6 +52,29 @@ List all domains for the account 1010 that have name matching "example":
...
+
+Name | Description
+-----|------------
+`:id` | Sort domains by ID
```
weppos: I'd remove the `:`, it just create confusions as it's not really a symbol.

weppos:
@jodosha to keep the docs simple, in the specific examples I'd just provide the sorting parameters.

Instead, we may want to add a small section called "Sorting" in `v2.markdown` where we explain how sorting work. In this way we avoid duplicating all the examples.

What do you think?

content/v2/domains.markdown
jacegu: What about:

> Multiple fields are allowed and _MUST_ be comma separated. The order of fields *IS* relevant, as it will determine which order criteria takes priority.

We can add an example with 3 tuples and 2 different orderings to show how the order affects the result.

content/v2/domains.markdown
jacegu: 👍

content/v2/domains.markdown
```diff
@@ -52,6 +52,29 @@ List all domains for the account 1010 that have name matching "example":
...
+`:name` | Sort domains by name
+`:expires_on` | Sort domains by expiration date
+`:created_at` | Sort domains by creation timestamp
+`:updated_at` | Sort domains by last update timestamp
+
+The default sorting policy is by ascending `:name`.
```
jacegu: We should probably add something like _"which is the equivalent to sorting them alphabetically"_. (Even though these docs are for technical people _by ascending `name`_ sounds way more confusing by alphabetically).

weppos: It seems appropriate:

```
+`:name` | Sort domains by name alphabetically
```

Actually, I wonder if it is alphabetically or lexicographical.

content/v2/domains.markdown
jodosha: @weppos I agree, but I added the `:` in the first place for consistency. If we plan to remove here, we should remove everywhere.

weppos: @jodosha the data in this table is different than the other places, for example params. In params I used `:account_id` because that is the name of the placeholder in the URI.

Here there is no placeholder at all, this is just a list of values.

content/v2/domains.markdown
jodosha: @weppos It's alphabetical, as it's the Postgres default.

jodosha:
@weppos @jacegu All the feedbacks were applied, can you please give this PR a second review? Thank you!

content/v2.markdown
```diff
@@ -157,6 +157,23 @@ For example:
 
     https://api.dnsimple.com/v2/_/domains?per_page=100&page=2
 
+## Sorting
+
+API v2 results are implictely sorted according to policies that vary from endpoint to endpoint.
```
jacegu: *implicitly*

content/v2.markdown
```diff
@@ -157,6 +157,23 @@ For example:
...
+
+You can decide your own sorting policy for each single API call via the `sort` param.
+This param accepts a pair of informations: the name of a field and the order (`asc` for ascending and `desc` for descending).
```
jacegu: _a pair of informations_ sounds odd to me. I'd probably go for something like _a key-value pair_. It also sounds like you can only pass one, when you can pass several. I'd tweak it to something like this:

> This param accepts a set of key-value pairs: the name of a field and the order criteria (`asc` for ascending and `desc` for descending).

jacegu:
@jodosha other than a typo and the tweaks for that sentence I say :shipit: (we may want our Chief Writing Director aka @jcaudle to give it a brief look and see if he has anything to add).

jcaudle:
I've provided a couple edits. What do you think @jodosha?

jodosha:
@jcaudle It looks good, thank you!

PR #73, opened 2016-05-13
v2 pagination: can return maximum 100 per page

3flex:
When I asked by email, was told that the maximum supported per_page value is 100.


weppos:
That's correct, thanks for the update.

PR #74, opened 2016-06-01
Remove support for wildcard account ID

jodosha:


content/v2.markdown
```diff
@@ -169,7 +169,7 @@ The order of fields is relevant, as it will determine the priority of the sortin
 
 ### Example
 
-List all domains for the account 1010 sorting them by expiration date (asc) and id (desc)
+List all domains from the account `1010` sorting them by expiration date (asc) and id (desc)
```
weppos: Is _from_ correct here?

/cc @aeden @jcaudle @arnamak

jcaudle: "from" is probably better than "for", but "in" seems like it'd be best.

jodosha:
@jcaudle Thank you 👍 
@weppos Do you think we can merge it now?

weppos:
👍

PR #75, opened 2016-06-07
Rename Contacts#email_address to Contacts#email

weppos:
All the clients should be updated.

/cc @jodosha @jacegu


jodosha:
@weppos 👍

PR #77, opened 2016-06-09
v2: correct name for record type filter

3flex:
When testing the endpoint `record_type` didn't work but `type` did.


jacegu:
Thanks for spotting this!

PR #78, opened 2016-06-10
API v2: Sorting part 2

jodosha:


jodosha:
@weppos @jacegu Please review, thanks! 😸

content/v2/tlds.markdown
```diff
@@ -47,6 +47,16 @@ List all TLDs.
 <%= pretty_print_fixture("/listTlds/success.http") %>
 ~~~
 
+### Sorting
+
+For general information about sorting, please refer to the [main guide](/v2/#sorting).
+
+Name | Description
+-----|------------
+`tld` | Sort TLDs by tld
```
jacegu: I know this is how we named the field but it sounds confusing... _Short the Foos by foo_.  Maybe _suffix_ onr _name_ could have been a better name.

jacegu:
@jodosha LGTM 👍

content/v2/tlds.markdown
weppos: Actually, as this is the only possible value, and there is no alternative, we may simply want to omit it. Same for webhooks.

weppos: Or explain what is the default order, and that's it.

jacegu: I think it's valuable as a way of knowing what's the ordering criteria to be expected from the endpoint. I'm 👍 on keeping it.

jodosha: @weppos @jacegu We can actually sort TLDs by `tld`.

I understand the _Sort Foos by foo_ thing, but in the payload we return a `tld` attribute: https://github.com/aetrion/dnsimple/blob/master/spec/api/v2/serializers/suffix_serializer_spec.rb#L10

PR #80, opened 2016-06-13
Cross-link the registrar and contacts endpoints

martinisoft:
This is to address some customer confusion about where to obtain this id. It's not consistent in name. I would expect contact_id here instead of registrant_id, but I get why it's named in context for this endpoint.

### DIFF
--- a/content/v2/registrar.markdown
+++ b/content/v2/registrar.markdown
@@ -77,9 +77,7 @@ Name | Type | Description
 `extended_attributes` | `hash` | **Required** for TLDs that require [extended attributes](/v2/tlds/#extended-attributes).
 
 <note>
-The `registrant_id` can be obtained via the [contacts endpoint](/v2/contacts) and will be the registered contact for this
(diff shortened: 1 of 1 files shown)


martinisoft:
Further clarified this with help from the other developers.

content/v2/registrar.markdown
```diff
@@ -76,6 +76,12 @@ Name | Type | Description
 `auto_renew` | `bool` | Set to true to enable the auto-renewal of the domain. Default: `true`.
 `extended_attributes` | `hash` | **Required** for TLDs that require [extended attributes](/v2/tlds/#extended-attributes).
 
+<note>
+The `registrant_id` can be obtained via the [contacts endpoint](/v2/contacts) and will be the registered contact for this
+domain. It is named differently due to the domain registration context technically allowing a registrant, technical, and
```
weppos: @martinisoft I'd skip the second part as this is not something we are planning anytime soon (from _It is named differently_).

weppos:
Merged, thanks!

PR #84, opened 2016-06-17
Update preview of ddns.sh for API v2 changes

tdurk93:
ddns.sh is now different from the preview on the ddns page. This left me confused for a minute. I updated the web page to reflect the new ddns.sh file.


aeden:
Thanks!

PR #85, opened 2016-06-17
Include account id in ddns.sh URL

tdurk93:
The URL in ddns.sh needs an account id to work. Currently, it returns a 404.


aeden:
Thank you again for the fixes on this, I appreciate it!

PR #86, opened 2016-06-18
Update ddns markdown to reflect updated script

tdurk93:
Sorry I should have updated this when I updated the ddns.sh script.


aeden:
Thanks for updating this as well.

PR #88, opened 2016-06-27
API v2: documentation for `applyTemplate`

jodosha:
@weppos @jacegu Please review. Thanks! 👍

### DIFF
--- a/content/v2/domains/templates.markdown
+++ b/content/v2/domains/templates.markdown
@@ -19,7 +19,7 @@ Name | Type | Description
 -----|------|------------
 `:account` | `integer` | The account id
 `:domain` | `string`, `integer` | The domain name or id
-`:template` | `integer` | The template id
+`:template` | `string`, `integer` | The template id or short name
 
 ### Example
 
--- a/content/v2/domains/templates.markdown
+++ b/content/v2/domains/templates.markdown
@@ -1,0 +1,37 @@
+---
+title: Templates API | Domains | DNSimple API v2 (Beta)
+excerpt: This page documents the DNSimple templates API v2.
+---
+
+# Templates API
+
+* TOC
+{:toc}
+
+
+## Apply a template {#apply}
+
+    POST /:account/domains/:domain/templates/:template
+
+### Parameters
+
+Name | Type | Description
+-----|------|------------
+`:account` | `integer` | The account id
+`:domain` | `string`, `integer` | The domain name or id
+`:template` | `integer` | The template id
+
+### Example
+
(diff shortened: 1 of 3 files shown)


content/v2/domains/templates.markdown
```diff
@@ -0,0 +1,37 @@
...
+
+    POST /:account/domains/:domain/templates/:template
+
+### Parameters
+
+Name | Type | Description
+-----|------|------------
+`:account` | `integer` | The account id
+`:domain` | `string`, `integer` | The domain name or id
+`:template` | `integer` | The template id
```
weppos: As far as I remember, the template ID can be either the ID or short name. Am I wrong?

jodosha: Right: 72398eb

content/v2/domains/templates.markdown
```diff
@@ -0,0 +1,37 @@
+---
+title: Templates API | Domains | DNSimple API v2 (Beta)
+excerpt: This page documents the DNSimple templates API v2.
```
jacegu: Being picky I'd say something like: _DNSimple API v2 template documentation_ (avoiding the _This page_ part which makes me uneasy).

weppos: @jacegu I believe this is a common description in all pages. If we make any change, it should be in all pages.

Also, it should look like a welcoming description and end with a dot. It is likely the description that search engines will display.

I don't think it's reasonable to make any change in this PR.

layouts/v2-sidebar.html
```diff
@@ -17,6 +17,7 @@ <h4 class="sidebar-title">API v2</h4>
   <li><ul class="nav-list">
     <li><a href="/v2/domains/certificates/">Certificates</a></li>
     <li><a href="/v2/domains/email-forwards/">Email Forwards</a></li>
+    <li><a href="/v2/domains/templates/">Templates</a></li>
```
jacegu: [There is](https://github.com/dnsimple/dnsimple-developer/blob/master/layouts/v2-sidebar.html#L37) already a Template link in the sidebar.

content/v2/domains/templates.markdown
jacegu: > I believe this is a common description in all pages. If we make any change, it should be in all pages.

It sucks but I'll 😶 then.

layouts/v2-sidebar.html
jodosha: @jacegu You're right. I moved them under Templates section. The structure now reflects Services > Domains.

PR #89, opened 2016-06-29
API v2: RSS feed for Changelog

jodosha:
It will be available at: https://developer.dnsimple.com/v2/feed.xml

---

/cc @weppos


jacegu:
👍

aeden:
:shipit:

PR #95, opened 2016-08-11
Api v2: Domains > Pushes

jodosha:
@weppos @jcaudle Please review. Thanks! 🐭


content/v2/domains/pushes.markdown
```diff
@@ -0,0 +1,159 @@
+---
+title: Pushes API | Domains | DNSimple API v2 (Beta)
+excerpt: This page documents the DNSimple pushes API v2.
+---
+
+# Pushes API
```
weppos: I know we use the plural, but I wonder if we should use "Push API" here.

content/v2/domains/pushes.markdown
```diff
@@ -0,0 +1,159 @@
...
+          -H 'Content-Type: application/json' \
+          -X POST \
+          -d '<json>' \
+          https://api.dnsimple.com/v2/2020/pushes/1
+
+### Input
+
+Name | Type | Description
+-----|------|------------
+`contact_id` | `integer` | **Required** - A contact that belongs to the target DNSimple account.
```
weppos: I'd expand it to:

> A contact that belongs to the target DNSimple account. The contact will be used as new registrant for the domain, if the domain is registered with DNSimple.

weppos:
:shipit:

PR #99, opened 2016-09-09
Fixture appliedServices should be listAppliedServices

jacegu:
I think `appliedServices` doesn't follow the convention.

@aeden @weppos 👍 👎 ?


aeden:
The only reason I would consider sticking with how it is, is because applied services aren't really CRUD resources, so we don't treat them like the other CRUD resources (with list, create, get, update and delete).

jacegu:
Won't fix.

PR #118, opened 2016-12-11
API v2: ISO-8601 timestamps

jodosha:


weppos:
@jodosha have you also added the paragraph in the documentation overview?

jodosha:
@weppos now we have 😉

content/v2.markdown
```diff
@@ -110,6 +110,17 @@ Error responses will be returned as a JSON object with at least a `message` key.
...
+## Timestamps
+
+All the timestamps and the dates are formatted according the [ISO-8601 standard](http://www.iso.org/iso/es/home/standards/iso8601.htm).
+
+~~~
+2016-12-11T17:31:51Z # Timestamp
+2016-12-11           # Date
+~~~
+
+All the timestamps are UTC.
```
weppos: I would actually encourage to extract the TZ from the date. If you want add this line, perhaps we can say:

> All the timestamps are UTC. However, you are encouraged to parse the time zone from the date.

PR #128, opened 2016-12-19
Clarify that web hooks might fire more then once

nesQuick:


content/v2/webhooks.markdown
```diff
@@ -15,7 +15,7 @@ Webhooks are used for receiving notifications via an HTTP request whenever an ev
 
 You may register your webhook URLs either through the DNSimple web interface or through the DNSimple API. The URL must use HTTPS and the webhook handler should be able to receive HTTP POST requests.
 
-All webhooks are sent via HTTPS using POST.
+All webhooks are sent via HTTPS using POST. It MAY occur that some webhook fire more then once, the `request_identifier` can be used to detect and handle these cases.
```
aeden: I'd go a step further and say that clients *MUST* handle the same webhook identifier received multiple times.

PR #131, opened 2016-12-20
Document how to get the account ID for account-scoped methods

weppos:
This was requested today from a customer. I realized we did not properly provide any instruction on how the account ID is supposed to be fetched.

### DIFF
--- a/content/v2.markdown
+++ b/content/v2.markdown
@@ -39,9 +39,9 @@ The current API version is `v2`.
 
 There two ways to authenticate through DNSimple API: the **HTTP Basic Authentication** and the **OAuth2 token**.
 
-When you create an OAuth2 token programmatically, the token is scoped to the specific account granted during the OAuth2 dance. When you create an OAuth2 token from the application, you can create either an user token or an account token. The user token gives you access to any resource associated to any account the user has access to. Conversely, an account token gives you access only to the resources connected to that account.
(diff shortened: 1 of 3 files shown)


content/v2.markdown
```diff
@@ -118,6 +125,46 @@ All the timestamps and the dates are formatted according the [ISO-8601 standard]
...
+
+##### Getting the ID programmatically with an Account access token
+
+If you use an account token and you want to get the ID programmatically, the simplest way is to perform a [`whoami`](/v2/identity/#whoami) API call. The response will contain the account identifier as part of the `account` payload.
+
+~~~json
+<%= pretty_print_fixture("/whoami/success-account.http") %>
+~~~
+
+##### Getting the ID programmatically with an Account access token
```
3flex: should be "with a User access token"?

3flex:
Small grammatical request - please use "a user" not "an user" throughout. If you say them both out loud the latter sounds odd. http://english.stackexchange.com/questions/105116/is-it-a-user-or-an-user

weppos:
@3flex thanks for the link. As a non-native English speaker, this is something quite hard to remember to me as I'm used to the simplest approach _vowel-based_.

I agree that it should be _a user_.

jacegu:
As I was going through this again I am wondering if we should have mentioned the `whoami` endpoint as a way of obtaining the `account_id`.

weppos:
@jacegu I did, see https://developer.dnsimple.com/v2/#account-scope

> If you use an account token and you want to get the ID programmatically, the simplest way is to perform a whoami API call. The response will contain the account identifier as part of the account payload.

Do you have something different in mind?

jacegu:
@weppos sorry, I missed that. I was expecting reading `whoami` in one the titles of the 3 ways of getting it and I totally over looked it. 😞

PR #134, opened 2016-12-21
Update nanoc

nesQuick:
Also updated the gems 💎

I'd love to have some review of the generated site, I don't know the heavy bits to check. It compiles ¯\_(ツ)_/¯


weppos:
Thanks @nesQuick ❤️ 

I've just tested it locally, and it all seems to work as expected.

PR #136, opened 2016-12-21
Update method overviews

weppos:
This PR updates the initial section of each method, making sure they all follow the same format.

I am also renaming the method anchors to match the names we used consistently in the clients and internally.


weppos:
Before I continue, any objection in the format @dnsimple/dnsimple ?

jacegu:
👍

jodosha:
👍

PR #137, opened 2016-12-21
clarify dnsimple sandbox

### LINKED ISSUE #122: Clarify sandbox data separation
Reading trough the Sandbox documentation I found the following statement:

> This site is generally an exact duplicate of the production application [...]

it would be helpful to call the user to create a account, and make it clear that the data is not synced in any way. (That was at least my first impression)

### LINKED ISSUE #129: Clarify sandbox documentation
The sandbox documentation at https://developer.dnsimple.com/sandbox/ is not clear on the purpose and value of the sandbox. This documentation needs to be updated to clarify why the sandbox is useful and how it should be used.

nesQuick:
PTAL @jacegu 

also @dnsimple/dnsimple 

fixes #122
fixes #129

### DIFF
--- a/content/sandbox.markdown
+++ b/content/sandbox.markdown
@@ -10,7 +10,7 @@ excerpt: Welcome to DNSimple. This page explains our sandbox environment. Enjoy 
 
 ---
 
-We maintain an isolated test site at [sandbox.dnsimple.com](https://sandbox.dnsimple.com/). Please feel invited to test your applications that integrate with the DNSimple API here.
+We maintain an isolated test site at [sandbox.dnsimple.com](https://sandbox.dnsimple.com/). We highly recommend testing against our [sandbox environment](https://developer.dnsimple.com/sandbox/) before using our production environment. This will allow you to avoid real purchases, live charges on your credit card, and reduce the chance of your running up against rate limits.
 
 ### Disclaimer
 
@@ -18,7 +18,11 @@ Sandbox data and systems are generally treated like production, however we may n
 
 ## Getting started
 
-The DNSimple sandbox is a independent system, you have to create another account and register some domains to test your integrations. *Every action within the Sandbox will be free of charge*
+The DNSimple sandbox is a independent system, you have to create another account and register some domains to test your integrations. 
+
+<callout>
+Every action within the Sandbox will be free of charge.
+</callout>
 
 - Go to [sandbox.dnsimple.com/signup](https://sandbox.dnsimple.com/signup) and register your account. _It does not have to match your DNSimple live account._
 - When asked to choose a plan choose `Skip this step and choose a plan later` at the bottom. If you want to test features that requires a valid subscription please refer to the [Testing Subscriptions](#testing-subscriptions) section.
@@ -36,11 +40,15 @@ This site is generally an exact duplicate of the production application, but the
 
 To register domains with your sandbox account you will need an active subscription ([see below](#testing-subscriptions)). You can register every domain you want, as long as it's available because you are still sharing a system with other customers.
 
-*Domains you register within the sandbox will not have any DNS service*
+<callout>
+Domains you register within the sandbox will not have any DNS service.
+</callout>
+
 
 ## Testing Certificates
 
 Testing working with certificates is at the moment *not supported* within the sandbox because certificate validation requires actual DNS service.
+
 
 ## Testing Subscriptions
 
@@ -50,6 +58,7 @@ When entering the credit card details:
 
 - Use the credit card number `1` to simulate a credit card in good status. All payments and authorizations will succeed.
 - Use the credit card number `3` to simulate a credit card in which all authorizations and payments will fail.
+
 
 ## Testing Two Factor Authentication
 
--- a/content/sandbox.markdown
+++ b/content/sandbox.markdown
@@ -10,20 +10,36 @@ excerpt: Welcome to DNSimple. This page explains our sandbox environment. Enjoy 
 
 ---
 
-We maintain an isolated test site at [sandbox.dnsimple.com](https://sandbox.dnsimple.com/). Please feel free to test your applications that integrate with the DNSimple API here.
+We maintain an isolated test site at [sandbox.dnsimple.com](https://sandbox.dnsimple.com/). Please feel invited to test your applications that integrate with the DNSimple API here.
+
+### Disclaimer
 
 Sandbox data and systems are generally treated like production, however we may not apply the same backup, monitoring and high availability policies. That means we may occasionally need to take this environment down, wipe out a portion, or clear all stored data.
 
+## Getting started
 
-## URL
+The DNSimple sandbox is a independent system, you have to create another account and register some domains to test your integrations.
+
+- Go to [sandbox.dnsimple.com/signup](https://sandbox.dnsimple.com/signup) and register your account. _It does not have to match your DNSimple live account._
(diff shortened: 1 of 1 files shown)


content/sandbox.markdown
```diff
@@ -10,20 +10,36 @@ excerpt: Welcome to DNSimple. This page explains our sandbox environment. Enjoy
...
-We maintain an isolated test site at [sandbox.dnsimple.com](https://sandbox.dnsimple.com/). Please feel free to test your applications that integrate with the DNSimple API here.
+We maintain an isolated test site at [sandbox.dnsimple.com](https://sandbox.dnsimple.com/). Please feel invited to test your applications that integrate with the DNSimple API here.
+
+### Disclaimer
 
 Sandbox data and systems are generally treated like production, however we may not apply the same backup, monitoring and high availability policies. That means we may occasionally need to take this environment down, wipe out a portion, or clear all stored data.
 
+## Getting started
+
+The DNSimple sandbox is a independent system, you have to create another account and register some domains to test your integrations.
```
jacegu: It may be worth mentioning that the account and domains will be free of charge so it doesn't scare anyone off at this point of the read.

content/sandbox.markdown
```diff
@@ -10,20 +10,36 @@ excerpt: Welcome to DNSimple. This page explains our sandbox environment. Enjoy
...
+### Sandboxing 3rd parties
+
+This site is generally an exact duplicate of the production application, but there are some exceptions. Instead of actually performing domain registration, charging credit cards, etc. it _mocks_ these functions.
+
+
+## Testing Domains
+
+To register domains with your sandbox account you will need an active subscription ([see below](#testing-subscriptions)). You can register every domain you want, as long as it's available because you are still sharing a system with other customers.
+
+*Domains you register within the sandbox will not resolve*
```
jacegu: I'd replace will not resolve:  *Domains you register within the sandbox will not have DNS service*

content/sandbox.markdown
```diff
@@ -10,20 +10,36 @@ excerpt: Welcome to DNSimple. This page explains our sandbox environment. Enjoy
...
 
-This site is generally an exact duplicate of the production application, but there are some exceptions. Instead of actually performing domain registration, charging credit cards, etc. it mocks these functions.
 
-### Testing Subscriptions
+## Testing Subscriptions
 
 To test mock registration or other subscription-based features you will have to create a test account.
```
jacegu: The first part of this sentence sounds odd to me. Probably *mock* is not needed there?

content/sandbox.markdown
```diff
@@ -10,20 +10,36 @@ excerpt: Welcome to DNSimple. This page explains our sandbox environment. Enjoy
...
 
 The API hostname is `api.sandbox.dnsimple.com`. This hostname follows the same [URL and version](/overview/#url) conventions of the production environment.
 
-## Special Settings
+### Sandboxing 3rd parties
+
+This site is generally an exact duplicate of the production application, but there are some exceptions. Instead of actually performing domain registration, charging credit cards, etc. it _mocks_ these functions.
+
+
+## Testing Domains
```
jacegu: We should probably add a similar section about certificates (as in *Testing Certificates*).

nesQuick: I try to figure out what the sandbox actually can do with certs and will add it then :)

nesQuick:
@jacegu fixed the minors, working on the cert section :)

nesQuick:
added a section about certificates. But I really want to have review of @weppos and @jacegu of that one

content/sandbox.markdown
```diff
@@ -10,29 +10,48 @@ excerpt: Welcome to DNSimple. This page explains our sandbox environment. Enjoy
...
+
+To register domains with your sandbox account you will need an active subscription ([see below](#testing-subscriptions)). You can register every domain you want, as long as it's available because you are still sharing a system with other customers.
+
+*Domains you register within the sandbox will not have DNS service*
```
jacegu: I was wondering if we should also mention other DNS related services like Secondary DNS. Thoughts?

nesQuick: Hmm, we could expend to "any DNS services". I have no experience how many people are actually using it, I think at some point we should decide where to make a cut, also in order to keep the docs concise and not to overwhelm people who come here for the first time.

jacegu: Sounds good 👍

content/sandbox.markdown
```diff
@@ -10,29 +10,48 @@ excerpt: Welcome to DNSimple. This page explains our sandbox environment. Enjoy
...
 
-## Special Settings
+### Sandboxing 3rd parties
+
+This site is generally an exact duplicate of the production application, but there are some exceptions. Instead of actually performing domain registration, charging credit cards, etc. it _mocks_ these functions.
+
+
+## Testing Domains
+
+To register domains with your sandbox account you will need an active subscription ([see below](#testing-subscriptions)). You can register every domain you want, as long as it's available because you are still sharing a system with other customers.
```
jacegu: Another thing that just came into my mind: should we mention that it is not possible to transfer domains to the sandbox environment?

nesQuick: I think this somehow relates to not having any DNS services. But we could also mark this explicit in here. You think we should?

content/sandbox.markdown
```diff
@@ -10,29 +10,57 @@ excerpt: Welcome to DNSimple. This page explains our sandbox environment. Enjoy
...
+To test domain registrations or other subscription-based features you will have to subscribe to a plan.
 
 When entering the credit card details:
 
 - Use the credit card number `1` to simulate a credit card in good status. All payments and authorizations will succeed.
 - Use the credit card number `3` to simulate a credit card in which all authorizations and payments will fail.
 
-### Testing Two Factor Authentication
+
+## Testing Two Factor Authentication
```
weppos: I think we need to revise this section. The `00000` token was when we used Authy. Since we switched to an OTP based system, then 2FA simply requires a properly configured device.

content/sandbox.markdown
```diff
@@ -10,29 +10,57 @@ excerpt: Welcome to DNSimple. This page explains our sandbox environment. Enjoy
...
-### Testing Subscriptions
+<callout>
+Domains you register within the sandbox will not have any DNS service.
+</callout>
 
-To test mock registration or other subscription-based features you will have to create a test account.
+
+## Testing Certificates
+
+Testing working with certificates is at the moment *not supported* within the sandbox because certificate validation requires actual DNS service.
```
weppos: I'm not completely convinced by this. It _should_ work, but sadly this is so unaffordable that I'm fine to say you can't.

nesQuick: I wonder how this is supposed to work. As I understood I need to click something in an email I've received on a domain which has no DNS service (MX record). Or how am I supposed to verify that I *own* the domain in question?

weppos: If you own the domain _in real life_ and you have MX configured, you will receive an approval email.

nesQuick:
@weppos I wonder if we should remove the 2FA part completely now. Because if you ignore history of this article it does not make sense to mention it as it functions as the rest of the system :)

weppos:
@nesQuick I'm fine if you want to remove it

nesQuick:
@weppos alright, I did delete it.

From my POV it's good to merge, any objections? @dnsimple/dnsimple

content/sandbox.markdown
nesQuick: True, I see the point. I'm totally with your first point then, to pretend it's not working. People  can find out :)

PR #141, opened 2017-01-16
v2 OAuth: Document additional required parameters

### LINKED ISSUE #140: OAuth: DNSimple requires undocumented "state" parameter in token request
According to RFC 6749 "state" is only used in authorization request and authorization response. However in the DNSimple implementation (in the sandbox at any rate) it's required at step 2 as a POST parameter. This isn't in compliance with the spec, and also isn't documented in the DNSimple docs.

3flex:
Fixes #140


3flex:
Based on your comment in https://github.com/dnsimple/dnsimple-ruby/pull/146 I realize that `redirect_uri` is actually only required when provided in the authorization request, it's not always required... I'll submit another PR to clarify that.

PR #142, opened 2017-02-20
Updates the API documentation with DNSSEC changes

aeden:
Adds two new webhooks: `dnssec.rotation_start` and `dnssec.rotation_complete`. These webhooks are used by clients with domains registered outside of DNSimple, and who want to enable DNSSEC on their zones.

### DIFF
--- a/content/v2/domains/dnssec.markdown
+++ b/content/v2/domains/dnssec.markdown
@@ -60,6 +60,8 @@ Responds with HTTP 204.
 <%= pretty_print_fixture("/disableDnssec/success.http") %>
 ~~~
 
+Responds with HTTP 428 if DNSSEC is not currently enabled.
+
 
 ## Get DNSSEC {#get}
 
--- a/content/v2/domains/dnssec.markdown
+++ b/content/v2/domains/dnssec.markdown
@@ -1,0 +1,228 @@
+---
+title: DNSSEC API | Domains | DNSimple API v2
+excerpt: This page documents the DNSimple DNSSEC API v2.
+---
+
+# DNSSEC API
+
+* TOC
+{:toc}
+
+
+## Enable DNSSEC {#enable}
+
+    POST /:account/domains/:domain/dnssec
+
+Enable DNSSEC for the domain in the account. This will sign the zone. If the domain is registered it will also add the DS record to the corresponding registry.
+
+### Example
+
+Enable DNSSEC for the domain `example.com` in the account `1010`:
+
+    curl  -H 'Authorization: Bearer <token>' \
+          -H 'Accept: application/json' \
+          -X POST \
+          https://api.dnsimple.com/v2/1010/domains/example.com/dnssec
+
+### Response
+
+Responds with HTTP 201.
+
+~~~json
+<%= pretty_print_fixture("/enableDnssec/success.http") %>
+~~~
+
+
+## Disable DNSSEC {#disable}
+
+    DELETE /:account/domains/:domain/dnssec
+
+### Example
+
+Disable DNSSEC for the domain `example.com` in the account `1010`:
+
+    curl  -H 'Authorization: Bearer <token>' \
+          -H 'Accept: application/json' \
+          -X DELETE \
+          https://api.dnsimple.com/v2/1010/domains/example.com/dnssec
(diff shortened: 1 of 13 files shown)


content/v2/domains/dnssec.markdown
```diff
@@ -0,0 +1,228 @@
...
+Disable DNSSEC for the domain `example.com` in the account `1010`:
+
+    curl  -H 'Authorization: Bearer <token>' \
+          -H 'Accept: application/json' \
+          -X DELETE \
+          https://api.dnsimple.com/v2/1010/domains/example.com/dnssec
+
+### Response
+
+Responds with HTTP 204.
```
weppos: Am I wrong, or we also respond with 412?

content/v2/domains/dnssec.markdown
```diff
@@ -0,0 +1,228 @@
...
+    curl  -H 'Authorization: Bearer <token>' \
+          -H 'Accept: application/json' \
+          https://api.dnsimple.com/v2/1010/domains/example.com/dnssec
+
+### Response
+
+Responds with HTTP 200.
+
+~~~json
+<%= pretty_print_fixture("/getDnssec/success.http") %>
```
weppos: Should we show both responses? one for enabled and one for disabled

content/v2/domains/dnssec.markdown
```diff
@@ -0,0 +1,228 @@
...
+title: DNSSEC API | Domains | DNSimple API v2
+excerpt: This page documents the DNSimple DNSSEC API v2.
+---
+
+# DNSSEC API
+
+* TOC
+{:toc}
+
+
```
weppos: We should add the "beta" banner. I'd also add a short intro that explains what DNSSEC is for, when you should use DNSSEC API vs DS Record API.

Nothing too complex, but to avoid devs would start sending DS record requests when they just need to use enable/disable DNSSEC.

content/v2/domains/dnssec.markdown
aeden: I believe 412 is returned if the feature is gated, correct?

428 is the other response, when the feature is not enabled. I will add that.

content/v2/domains/dnssec.markdown
aeden: Nah, it's basically the same thing, one with a true, one with a false.

content/v2/domains/dnssec.markdown
aeden: I've added the beta banner and a note in the create delegation signer endpoint regarding when it should be used.

content/v2/domains/dnssec.markdown
```diff
@@ -0,0 +1,238 @@
...
+### Input
+
+Name | Type | Description
+-----|------|------------
+`algorithm` | `string` | **Required**.
+`digest` | `string` | **Required**.
+`digest_type` | `string` | **Required**.
+`keytag` | `string` | **Required**.
+
```
jodosha: We should document `algorithm`, and `digest_type` with an explanation of these numbers, like we did here: https://developer.dnsimple.com/v2/tlds/#tld-types

content/v2/domains/dnssec.markdown
```diff
@@ -0,0 +1,238 @@
...
+* TOC
+{:toc}
+
+
+## Enable DNSSEC {#enable}
+
+    POST /:account/domains/:domain/dnssec
+
+Enable DNSSEC for the domain in the account. This will sign the zone. If the domain is registered it will also add the DS record to the corresponding registry.
+
```
aeden: There are no parameters for DNSSEC enable, disable, and get.

jodosha: I mean Parameters, not Input. See https://github.com/dnsimple/dnsimple-developer/pull/142/files#diff-12c45f6c13ebeebc73fec6c001707b85R95

aeden: Ah, thanks for the clarification. Adding.

content/v2/domains/dnssec.markdown
aeden: Done.

PR #146, opened 2017-05-03
Fixes a small typo in Webhooks documentation

benedikt:


jacegu:
Thanks for the fix @benedikt

PR #153, opened 2017-10-30
API v2: documentation for getZoneDistribution

jodosha:


content/v2/zones.markdown
```diff
@@ -117,3 +117,44 @@ Responds with HTTP 200, renders the zone file.
...
+<%= pretty_print_fixture("/getZoneDistribution/success.http") %>
+~~~
+
+Responds with HTTP 200 when the zone is not distributed.
+
+~~~json
+<%= pretty_print_fixture("/getZoneDistribution/failure.http") %>
+~~~
+
+Responds with HTTP 500 when the server failed to perform the check.
```
jacegu: This sounds scary (`500` is a number that always is).

Not sure how often we should expect this to happen but I'd clearly explain it or rephrase this to clearly state that it's an error or unexpected behaviour.

content/v2/zones.markdown
```diff
@@ -117,3 +117,44 @@ Responds with HTTP 200, renders the zone file.
 <%= pretty_print_fixture("/getZoneFile/success.http") %>
 ~~~
 
+
+## Get zone distribution {#get-zone-distribution}
+
+    GET /:account/zones/:zone
+
+Checks if a zone is fully distributed across DNSimple nodes.
```
jacegu: Do you think it's worth it to explain what `fully distributed` means (or linking a corresponding support article, if any)?

content/v2/zones.markdown
jodosha: We don't expect it very often. This kind of error is returned in case the API server fails to contact all the nodes of our DNS cluster.

content/v2/zones.markdown
OleMchls: I second that thought, looking at this from customers :eyes: it might be pretty confusing. As it implies some internal (special) knowledge.

Maybe rephrasing is enough, coming up with a full explanation might be overkill and unnecessary for most customers. something along the lines of:

```
Checks if a zone change is fully distributed to all our nameservers in all different regions we support.
```

Maybe link in there to the regions docs.

jacegu:
Random question that just came to mind: is this something customers can test in Sandbox?

aeden:
@jacegu this feature is not testable with a sandbox domain, as those domains do not resolve.

jacegu:
Then we should clearly state this on the developer documentation.

content/v2/zones.markdown
weppos: I agree with @jacegu. 500 is the generic error when something goes wrong. This is for example the error we return when our system crashed.

In this case, we need to return a different error that makes it clear that it's not the API crashing, but it's the remote system that did not respond.

502 seems to be the most appropriate error. Sadly it's also the error returned by Nginx when not being able to connect to our backend app, but I'm not sure there is a better error.

content/v2/zones.markdown
```diff
@@ -117,3 +117,48 @@ Responds with HTTP 200, renders the zone file.
 <%= pretty_print_fixture("/getZoneFile/success.http") %>
 ~~~
 
+
+## Get zone distribution {#get-zone-distribution}
```
weppos: I would keep the feature in beta for at least a few months, so that we can have breaking changes. I would add a note at the beginning (warning) that this feature is currently considered beta and breaking changes may happen.

content/v2/zones.markdown
```diff
@@ -117,3 +117,48 @@ Responds with HTTP 200, renders the zone file.
...
+
+    GET /:account/zones/:zone
+
+Checks if a zone change is fully distributed to all our nameservers across the globe.
+
+<warning>
```
weppos: This is probably a note and not a warning.

content/v2/zones.markdown
```diff
@@ -117,3 +117,48 @@ Responds with HTTP 200, renders the zone file.
 <%= pretty_print_fixture("/getZoneFile/success.http") %>
 ~~~
 
+
+## Get zone distribution {#checkZoneDistribution}
```
weppos: I would keep the feature in beta for at least a few months, so that we can have breaking changes. I would add a note at the beginning (warning) that this feature is currently considered beta and breaking changes may happen.

weppos:
FYI I made a change in 0f7679c to use the same internal name, see #136

aeden:
@jodosha could you please get implement the requested change from @weppos and deploy. Thank you.

weppos:
@jodosha there is a specific template for that, see

    <%= render "v2-preview" %>

in https://github.com/dnsimple/dnsimple-developer/blob/master/content/v2/certificates.markdown

PR #155, opened 2017-12-28
Deprecate and remove premium_price attribute from registrar order responses

weppos:
This is inconsistent with most of our code, and if we will ever return a price it will be returned for every order.

## Clients

- [x] dnsimple/dnsimple-ruby#163
- [x] dnsimple/dnsimple-go#67
- [x] dnsimple/dnsimple-elixir#120
- ~dnsimple/dnsimple-node~ (not required)
- [x] dnsimple/dnsimple-java#10

### DIFF
--- a/content/v2/openapi.yml
+++ b/content/v2/openapi.yml
@@ -2548,9 +2548,6 @@ components:
           type: boolean
         whois_privacy:
           type: boolean
-        premium_price:
-          type: string
-          nullable: true
         created_at:
           $ref: '#/components/schemas/DatetimeISO8601'
         updated_at:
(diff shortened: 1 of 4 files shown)


weppos:
Going to update the clients once this PR is approved.

content/v2/openapi.yml
```diff
@@ -4112,7 +4100,7 @@ components:
             properties:
               period:
                 type: integer
-              preimum_price:
+              premium_price:
```
aeden: Isn't premium_price supposed to be removed?

weppos: Nope, this is the parameter. Not the response field.

aeden: Ah, ok. 👍

PR #157, opened 2018-01-12
be more explicit about the creditcard details when choosing a plan

fkr:


weppos:
Hi @fkr, thanks for the update. Can I ask you to describe what was confusing to you?

I am trying to interpret the change, it's not really clear to me the meaning of the new sentence. Perhaps I can merge it into the previous one and clarify it, but before I'd like to have a better sense of what confused you. :)

fkr:
Well, I had a support call regarding this with dnsimple and I simply had that support call because I found it to be not explicit enough that the '1' and '0' can be used during the sign-up to the sandbox as well.
As I understood it, if I test via the API I can use '0' and '1' for credit card, so I wanted to make sure that it is very explicit.

aeden:
Thank you for the feedback. I've merged the important information from the sentence into the previous sentence. Hopefully it is just as clear.

PR #152, opened 2017-10-19
API v2: document Let's Encrypt certificates

jodosha:
Documentation for the Let's Encrypt API endpoints: purchase/issue, renew/issue.

## Clients

- [x] dnsimple/dnsimple-ruby#159
- [x] dnsimple/dnsimple-go#63
- [x] dnsimple/dnsimple-elixir#118
- [x] dnsimple/dnsimple-node#14
- [x] dnsimple/dnsimple-java#8

PR #162, opened 2018-03-16
Fix Changelog header to reflect that API v2 is no longer in beta

woods:
I found one more spot that needed to be updated for v2 being stable. 

Let me know if you do prefer the plural in the second paragraph. I followed the conjugation from the announcement emails.

### DIFF
--- a/content/v2/changelog.markdown
+++ b/content/v2/changelog.markdown
@@ -5,9 +5,9 @@ excerpt: History of the changes to the API v2.
 
 # Changelog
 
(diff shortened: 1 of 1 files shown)


woods:
Along the same lines, are the Let's Encrypt endpoints officially out of beta now that the v2 is stable?

https://developer.dnsimple.com/v2/certificates/#purchaseLetsencryptCertificate

Here's a reference to a relevant historical conversation: https://github.com/dnsimple/dnsimple-developer/pull/152#pullrequestreview-79614213

weppos:
Thanks Scott!

PR #174, opened 2018-10-08
Move filters at the top, along with parameters

weppos:
This change makes filters and sorters more visible. Placing them after the response is not giving them the right importance.

This is the result of a direct customer feedback. Actually, we had a few reports in the last year as well.

### DIFF
--- a/content/v2/domains.markdown
+++ b/content/v2/domains.markdown
@@ -9,9 +9,11 @@ excerpt: This page documents the DNSimple domains API v2.
 {:toc}
 
 
-## List domains {#list}
+## List domains {#listDomains}
 
-    GET /:account/domains
+~~~
+GET /:account/domains
+~~~
 
 List domains in the account.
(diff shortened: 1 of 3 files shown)


content/v2/domains.markdown
```diff
@@ -9,9 +9,11 @@ excerpt: This page documents the DNSimple domains API v2.
 {:toc}
 
 
-## List domains {#list}
+## List domains {#listDomains}
```
aeden: I see you changed this, but only this instance in each file. Are you planning on changing the other instances of this in another PR (i.e. change from #create to #createDomain)?

weppos: Eventually, yes. But probably not part of this PR. This is a cleanup that will have to be made, we haven't committed resources to it yet.

I'm slowly making the changes as long as I'm working on any affected endpoint.

PR #176, opened 2018-10-29
Update information about test credit cards numbers

jacegu:
Since the old ones will no longer work.


jacegu:
> @jacegu can you please expand the section explaining we use Stripe, and link to the official Stripe docs as well?

Sure. Wasn't sure we wanted to disclose this information, so I didn't say anything. I'll ping you again once I update the PR.

PR #179, opened 2018-11-16
Let's Encrypt wildcard changes

OleMchls:
Once this feature goes into public beta this PR documents the API usage.

### DIFF
--- a/content/v2/certificates.markdown
+++ b/content/v2/certificates.markdown
@@ -180,7 +180,7 @@ Alternate names require a subscription to a Professional or Business plan.
 To request a wildcard certificate that's valid for an unlimited number of names that belong to a single subdomain level, use `*` (e.g. `*.example.com`).
 
 <info>
-Let's Encryot wildcard certificates is a feature that is only available to the following [new plans](https://support.dnsimple.com/articles/new-plans/#newer-plans-some): Professional or Business.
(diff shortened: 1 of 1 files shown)


content/v2/certificates.markdown
```diff
@@ -169,19 +169,25 @@ Custom names require a subscription to a Professional or Business plan.
...
-By default a certificate does not have alternate names.
+By default, a certificate doesn't have alternate names.
 
 You can purchase a single certificate for both `https://docs.example.com` and `https://status.example.com`, alongside  `https://example.com`.
 
 Alternate names require a subscription to a Professional or Business plan.
 
+### Wildcard certificates
+
+To request a wildcard certificate that's valid for all subdomains of a given domain, use `*` (e.g. `*.example.com`).
```
weppos: @OleMchls please update this line to match the same clarification regarding the scope of the wildcard I suggested also in support.

content/v2/certificates.markdown
```diff
@@ -169,19 +169,25 @@ Custom names require a subscription to a Professional or Business plan.
...
+
+Wildcard certificates require a subscription to a Professional or Business plan.
```
weppos: Can we use here the same info format of https://github.com/dnsimple/dnsimple-developer/blob/master/content/v2/vanity.markdown#vanity-name-servers-api ?

PR #181, opened 2019-01-07
WHOIS Privacy Renewal (renewWhoisPrivacy)

jodosha:


content/v2/registrar/whois-privacy.markdown
```diff
@@ -111,3 +111,37 @@ Responds with HTTP 200 if WHOIS privacy is disabled.
 ~~~json
 <%= pretty_print_fixture("/api/disableWhoisPrivacy/success.http") %>
 ~~~
+
+## Renew whois privacy {#renew}
```
weppos: The anchor should be the same name we use internally.

jodosha: @weppos I'm confused. Do you mean it should have `renewWhoisPrivacy`? The rest of the anchors in this file don't use the internal naming, but generic naming like `#enable` or `#disable`.

weppos: Yes, that's the old style. Since at some point become confusing, I started switching them to the qualified identifiers we use, eg https://developer.dnsimple.com/v2/domains/#listDomains

It's a work in progress. I haven't dedicated too much time on it. I'm fine to leave renew unless we switch the whole page.

PR #184, opened 2019-02-25
add minimum_registration to components/schemas/TLD

jordane:
This was missing from the schema, but exists in the example, and is actually
used in the sandbox and prod api.


weppos:
@aeden I'm removing myself from review. This field was published almost unintentionally. 

We'll have to evaluate how to expose the various TLD settings, but we made some internal changes to the TLD structure and more has to come. I'm not ready to add this change for now.

sbastn:
Closing due to inactivity

PR #210, opened 2020-01-13
Include link to example payloads

jacegu:
Since we maintain a collection of fixtures for each webhook event and these are a valuable resource for customers using our API I think it's worth it to link them from the corresponding docs.

### DIFF
--- a/content/v2/webhooks.markdown
+++ b/content/v2/webhooks.markdown
@@ -132,7 +132,7 @@ The following events are available:
 
 ### Event payloads
 
-We maintain a repository of example payloads that you can use [here](https://github.com/dnsimple/dnsimple-developer/tree/master/fixtures/v2/webhooks).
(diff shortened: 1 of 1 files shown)


content/v2/webhooks.markdown
```diff
@@ -130,6 +130,9 @@ The following events are available:
 - zone\_record.delete
 - zone\_record.update
 
+### Event payloads
+
+We maintain a repository of example payloads that you can use [here](https://github.com/dnsimple/dnsimple-developer/tree/master/fixtures/v2/webhooks).
```
weppos:
```suggestion
We maintain a [repository of example payloads](https://github.com/dnsimple/dnsimple-developer/tree/master/fixtures/v2/webhooks) that you can use to design and test your integrations.
```

itsalyse:
> Works for me. I proposed a small change, the existing sentence seemed a bit convoluted to read to me. But it's personal opinion, not even an 🇺🇸 one. 😄

🇺🇸checking in to say it was a good suggestion 👍

jacegu:
Thanks for the feedback <3

PR #212, opened 2020-01-18
Add note about FQDN when creating a record

OleMchls:
We've never clearly stated that you do not have to send the FQDN here.

### DIFF
--- a/content/v2/zones/records.markdown
+++ b/content/v2/zones/records.markdown
@@ -129,7 +129,7 @@ curl  -H 'Authorization: Bearer <token>' \
 
 Name | Type | Description
 -----|------|------------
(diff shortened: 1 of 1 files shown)


content/v2/zones/records.markdown
```diff
@@ -129,7 +129,7 @@ curl  -H 'Authorization: Bearer <token>' \
 
 Name | Type | Description
 -----|------|------------
-`name` | `string` | **Required**. Use an empty string to create a record for the apex.
+`name` | `string` | **Required**. Your domain will be appended. Use an empty string to create a record for the apex.
```
weppos:
```suggestion
`name` | `string` | **Required**. The record name, without the domain. The domain will be automatically appended. Use an empty string to create a record for the apex. 
```

PR #216, opened 2020-02-05
2.6.5 breaks the build

dallasread:
As per https://github.com/laurilehmijoki/s3_website/issues/323 and https://github.com/dnsimple/dnsimple-support/commits/master/.ruby-version


dallasread:
@weppos Thanks. I think this was added from the force. I'll clean it up.

PR #224, opened 2020-02-21
Add domain states enum to openapi YAML

aeden:
The list of domain states was not present in the Domain schema. This adds the states.


weppos:
Thanks!

PR #228, opened 2020-03-03
Switch to asdf

sbastn:
Configures repository to use asdf: https://asdf-vm.com/


sbastn:
👍 Kept same ruby version.

PR #243, opened 2020-04-28
Clarify webhook delivery

### LINKED ISSUE #127: Clarify web hook reliability
Currently the docs say:

```
Responding to Webhooks

To confirm receipt of a webhook, your server must respond with an HTTP 200 response. Any other response will be considered an error and may cause the event to be retried.
```

I think it would be more precise if a web hook is retried or not. _may_ may be problematic ;)

OleMchls:
fixes #127

### DIFF
--- a/content/v2/webhooks.markdown
+++ b/content/v2/webhooks.markdown
@@ -59,9 +59,9 @@ The `actor` is an object describing the entity that triggered the event. This ma
 
 ### Responding to Webhooks
 
-To confirm receipt of a webhook, your server must respond with an HTTP 200 response. Any other response will be considered an error and will cause the deliviery to be retried.
+To confirm receipt of a webhook, your server must respond with an HTTP 200 response. Any other response will be considered an error and will cause the deliviery to be re-tried.
 
-_Note: The number of retries is limited, we will stop the delivery attempts after a while._
+_Note: The number of re-tries is limited. We will stop delivery attempts after a while._
 
 ### Event List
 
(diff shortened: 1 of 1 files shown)


content/v2/webhooks.markdown
```diff
@@ -59,7 +59,9 @@ The `actor` is an object describing the entity that triggered the event. This ma
 
 ### Responding to Webhooks
 
-To confirm receipt of a webhook, your server must respond with an HTTP 200 response. Any other response will be considered an error and may cause the event to be retried.
+To confirm receipt of a webhook, your server must respond with an HTTP 200 response. Any other response will be considered an error and will cause the deliviery to be re-tried.
+
+_Note: The number of re-tries is limited. We will stop delivery attempts after a while._
```
itsalyse, line 64: > after a while.

is there a hard number of retries or a set amount of time? It's always good to be as specific as possible with stuff like this.

kayla-broussard: This would also be my comment - I'd want to know how many delivery attempts are made / what time they would stop at.

OleMchls: There is a hard cap, yes. However, IDK if we want to expose this number to the public. That'd make it part of our public API. Thoughts @weppos @jacegu  ?

weppos: There's nothing secrets in the number of retries and we can share it, but it may also be subject to changes so I'm not sure it's worth to be as precise as saying 5, 10, 50, etc.

I think that if we remove:

> _Note: The number of re-tries is limited. We will stop delivery attempts after a while

And simply leave the change that was made (note there is a typo):

> To confirm receipt of a webhook, your server must respond with an HTTP 200 response. Any other response will be considered an error and will cause the **deliviery** to be re-tried.

we're good to go.

PR #248, opened 2020-05-16
Replace cancelDomainTransfer and getDomainTransfer fixtures

duduribeiro:
getDomainTransfer and cancelDomainTransfer fixtures are not in the same
pattern as the other fixtures. This commit replaces them to be in the
same format as the others.

### DIFF
--- a/lib/default.rb
+++ b/lib/default.rb
@@ -1,8 +1,6 @@
 require 'net/http'
 require 'json'
 
-
-
 include Nanoc::Helpers::Rendering
 include Nanoc::Helpers::XMLSitemap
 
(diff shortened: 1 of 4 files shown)


lib/default.rb
```diff
@@ -4,6 +4,19 @@
 include Nanoc::Helpers::Rendering
 include Nanoc::Helpers::XMLSitemap
 
+# TODO: Remove this monkey-patch if https://github.com/ruby/ruby/pull/3117 get merged.
```
duduribeiro, line 7: This was necessary because net/http fails if the status line is in the HTTP/2 format. This PR is open and we can remove this monkey-patch when (and if) it gets merged.

weppos: I suggest you use `--http1.1`. I suspect not all programming languages will be able to read HTTP/2 fixtures.
https://ec.haxx.se/http/http-versions

duduribeiro: updated!

PR #250, opened 2020-05-22
Fix Transfer-Encoding header in API fixtures

### LINKED ISSUE #247: Wrong `Transfer-Encoding` header in fixture files
Our servers produce responses with `Transfer-Encoding: chunked`, which means:
- Response don't have a `Content-Length` header
- Payload data is divided into several chunks
- Each chunk is preceeded by their content length in hexadecimal
- The end of the payload is signaled by a `0` content length followed by an empty line

  (source https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding)

This is an example of a response from our sandbox API showing these things:
```
HTTP/1.1 200 OK
Server: nginx
Date: Thu, 14 May 2020 08:44:52 GMT
Content-Type: application/json; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
X-RateLimit-Limit: 2400
X-RateLimit-Remaining: 2393
X-RateLimit-Reset: 1589448564
ETag: W/"1714b27c2c63e75c397413138fe869f0"
Cache-Control: max-age=0, private, must-revalidate
X-Request-Id: 689926a4-9279-4186-90d5-b3ba8391968a
X-Runtime: 0.006271
X-Frame-Options: DENY
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
X-Download-Options: noopen
X-Permitted-Cross-Domain-Policies: none
Strict-Transport-Security: max-age=31536000

af
{"data":[{"id":1324,"email":"guillermo.almazor@dnsimple.com","plan_identifier":"business-v1-monthly","created_at":"2020-03-16T16:10:53Z","updated_at":"2020-05-07T07:20:36Z"}]}
0

```

Our fixture set includes 118 fixtures that declare `Transfer-Encoding: chunked` but they don't actually declare any chunk in their payloads (there are even instances of empty payloads) which would make the fixture wrong in two ways:
1. It's not the actual exact response our API would produce.
2. HTTP clients won't be able to parse their payloads when replayed.

I think that we should amend this situation either by replacing all fixtures or changing the existing ones to `Transfer-Encoding: identity`.

### Replacing all fixtures

The following script will write in the standard output the full raw HTTP response:
```bash
#!/bin/bash

headers_file=$(mktemp /tmp/headers.XXXXXX)
payload_file=$(mktemp /tmp/payload.XXXXXX)

curl --raw \
     -s \
     -D $headers_file \
     -o $payload_file \
     -H "Authorization: Bearer $1" \
     -H 'Content-Type: application/json' \
     -X GET \
     $2

cat $headers_file
cat $payload_file

rm $headers_file
rm $payload_file
```
(This script only handles GET requests. It should be adapted to handle the other methods we need and to send payloads to the API endpoints)

Example usage: `$ ./fetch_fixture.sh YOUR_TOKEN_HERE https://api.sandbox.dnsimple.com/v2/accounts`

The tradeoff of doing this is that we will have to adapt the test suites of the clients like Java that have mocked HTTP clients that don't expect chunked payloads.

### Adapting all fixtures

A massive search and replace from `Transfer-Encoding: chunked` to `Transfer-Encoding: identity` will do the trick.

The tradeoff is that we are not providing fixtures that are exactly what our API produces.

ggalmazor:
Fixes #247 

This PR replaces the `Transfer-Encoding` header's value in all API fixtures from `chunked` to `identity` to make them replayable without further modification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,3 +47,10 @@ Finally, run:
 ```shell
 rake publish
 ```
+
+## Creating new fixtures
+
+- Use the sandbox API. Avoid using the production API as much as possible.
+- Run `bin/download_fixture.sh YOUR_TOKEN HTTP_METHOD URL > fixtures/v2/api/OPERATION/meaningful_name.http`
+
+  Example: `bin/download_fixture.sh 0123456789abcdefghijklmnopqrstuv GET https://api.sandbox.dnsimple.com/v2/tlds > fixtures/v2/api/listTlds/success.http`
--- a/bin/download_fixture.sh
+++ b/bin/download_fixture.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+curl -s \
+  -i \
+  --http1.1 \
+  -H "Authorization: Bearer $1" \
+  -H 'Content-Type: application/json' \
+  -H 'Accept: application/json' \
+  -X $2 \
+  $3 |
+  sed 's/^Transfer-Encoding: chunked/Transfer-Encoding: identity/'
--- a/fixtures/v2/api/accounts/success-account.http
+++ b/fixtures/v2/api/accounts/success-account.http
@@ -2,7 +2,7 @@ HTTP/1.1 200 OK
 Server: nginx
 Date: Tue, 14 Jun 2016 12:02:58 GMT
 Content-Type: application/json; charset=utf-8
-Transfer-Encoding: chunked
+Transfer-Encoding: identity
 Connection: keep-alive
 X-RateLimit-Limit: 2400
 X-RateLimit-Remaining: 2391
--- a/fixtures/v2/api/accounts/success-user.http
+++ b/fixtures/v2/api/accounts/success-user.http
@@ -2,7 +2,7 @@ HTTP/1.1 200 OK
 Server: nginx
 Date: Tue, 14 Jun 2016 12:05:38 GMT
 Content-Type: application/json; charset=utf-8
-Transfer-Encoding: chunked
+Transfer-Encoding: identity
 Connection: keep-alive
 X-RateLimit-Limit: 2400
 X-RateLimit-Remaining: 2390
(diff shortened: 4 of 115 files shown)


weppos:
@ggalmazor can you please prepare a simple shell script that wraps `curl -i -s --http1.1`, perform the request, and changes chunked to identity? It seems a nice addition to add in the fixtures folder.

In this way we limit issues around custom generation strategies.

weppos:
Before shipping, it is worth to take the ones for zone (as instance), and replace the in the other clients we have. Run the tests, and see if this change breaks anything.

ggalmazor:
@weppos, regarding the script you're asking me to create, I wrote and tested one that produces the fixtures we need (It's on #247):

```
#!/bin/bash

headers_file=$(mktemp /tmp/headers.XXXXXX)
payload_file=$(mktemp /tmp/payload.XXXXXX)

curl --raw \
     -s \
     -D $headers_file \
     -o $payload_file \
     -H "Authorization: Bearer $1" \
     -H 'Content-Type: application/json' \
     -X GET \
     $2

cat $headers_file
cat $payload_file

rm $headers_file
rm $payload_file
```

Provided I add `--http1.1`, would this be what you're asking for?

weppos:
If I run this script, will it produce a file like the ones you added to this repo?

ggalmazor:
> If I run this script, will it produce a file like the ones you added to this repo?

No, it would produce the ones that are exactly what our API returns, which have `Transfer-Encoding: chunked` and proper chunk definitions in the payload.

The ones I've changed in this PR effectively have the same headers (except `Transfer-encoding`) and payloads but are formally different.

weppos:
Please adjust the script to produce a fixture similar to what you have in this PR, otherwise it makes no sense to have the script.

ggalmazor:
@weppos, I've created the script at `bin/download_fixture.sh` and I've added a new block at the end of the `README.md` with instructions.

ggalmazor:
Replacing the fixtures in the java client with the ones from this PR will require just a couple of adjustments in the actual values we are verifying in 3 tests. Otherwise, it's a clean change.

PR #251, opened 2020-05-25
Remove required from Fax attribute

duduribeiro:
Working on the contacts API, I noticed that the docs are showing that
the fax is required but it is not.

This commit remove the required flag from it.

### DIFF
--- a/content/v2/contacts.markdown
+++ b/content/v2/contacts.markdown
@@ -101,7 +101,7 @@ Name | Type | Description
 `country`           | `string` | **Required**.
 `email`             | `string` | **Required**.
 `phone`             | `string` | **Required**.
-`fax`               | `string` | **Required**.
+`fax`               | `string` |
 
 ##### Example
 


weppos:
Assigning to another team member. I don't think I need to be a reviewer here.

Please make sure you all check the OpenAPI definition as well.

duduribeiro:
OpenAPI is right. It is marked as nullable.

ggalmazor:
@duduribeiro, you said that it's marked as nullable in OpenAPI, but I can't see `nullable: true`  in that field at https://github.com/dnsimple/dnsimple-developer/blob/master/content/v2/openapi.yml#L2691 

Maybe I'm looking at the wrong place?

duduribeiro:
> @duduribeiro, you said that it's marked as nullable in OpenAPI, but I can't see `nullable: true` in that field at https://github.com/dnsimple/dnsimple-developer/blob/master/content/v2/openapi.yml#L2691
> 
> Maybe I'm looking at the wrong place?

hey @ggalmazor, 👋 

this is the response one. The parameters are bellow `requestBodies`. This is the `fax` for parameter: https://github.com/dnsimple/dnsimple-developer/blob/master/content/v2/openapi.yml#L4370

PR #252, opened 2020-05-29
Removes references for Domain expires_on attribute

duduribeiro:
This removes the references of the deprecated Domain expires_on in
favor of expires_at attribute.


content/v2.markdown
```diff
@@ -271,7 +271,7 @@ List all domains in the account `1010` sorting them by expiration date (asc) and
 
     curl  -H 'Authorization: Bearer <token>' \
           -H 'Accept: application/json' \
-          https://api.dnsimple.com/v2/1010/domains?sort=expires_on:asc,id:desc
```
weppos, line 274: This is a breaking change. Therefore, we'll have to alias `expires_on` in our app to continue to work.

Our API should not be tied to the internal of our system. We used "modification", we should have used "expiration" here. I suggest you change the filter to `expiration`, and in our app we'll both handle `expires_on` and `expiration`.

duduribeiro: `expires_on` and `expires_at` is both working as a filter, but I agree with you that `expiration` is a better name for the filter

duduribeiro:
> I just realized this changeset doesn't include fixture updates. We'll have to update all the fixtures that represent a domain to make sure we add the new field.

 makes sense. I would do as next step, but agree that makes sense to have in this one

duduribeiro:
@weppos updated the api and webhook fixtures

PR #285, opened 2020-12-09
Add the first use-case tutorial on how to register a domain

DXTimer:
This PR introduces a short tutorial on how our customers might leverage our API to register domains on behalf of their customers. Feedback welcomed.


aeden (approved):
As a first pass I think this is good. I have a few text changes. I think we may also want to iterate on the use case once it is live.

itsalyse (approved):

PR #319, opened 2021-02-17
Create ifstated.conf

pguizeline:
Hi!

The main purpose of this ifstated.conf is to keep DNSimple API queries to a minimum, instead of calling ddns.sh every 5min via cron. The ddns.sh is still needed as quickly stated on the comments. The way it's implemented now is that ifstated runs ddns.sh only if it detects that the ping tests have failed.

It's a dirty hack, I'm sure more functionality can be added if needed. For example detecting quick dynamic DNS changes enforced by the ISP. Maybe using dig on special domains and comparing the IP against ifconfig outputs.

Let me know if further changes are needed! Thanks!

### DIFF
--- a/content/ddns/ifstated.conf
+++ b/content/ddns/ifstated.conf
@@ -1,0 +1,71 @@
+# Tested on OpenBSD 6.8
+
+# Based on https://github.com/vedetta-com/vedetta/blob/master/src/etc/ifstated.conf
+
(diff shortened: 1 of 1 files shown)


weppos:
Hi @pguizeline, thanks for the contributions. We'd rather not add more scripts/examples in this repo, as once merged we'll set expectations about maintenance.

If you want to add this script to a gist/repo you manage, I'll be happy to consider linking it in the tools section.
https://developer.dnsimple.com/tools/

PR #328, opened 2021-03-18
Add documentation for Secondary DNS API

duduribeiro:
This commit adds documentation for secondary DNS api endpoints. For
primary serves and secondary zones. It adds the information in the
openapi spec as well


layouts/sidebar.html
```diff
@@ -33,6 +33,7 @@ <h4 class="sidebar-title">API v2</h4>
     <li><a href="/v2/zones/records/">Records</a></li>
   </ul></li>
   <li><a href="/v2/contacts/">Contacts</a></li>
+  <li><a href="/v2/secondary-dns/">Secondary DNS</a></li>
```
DXTimer, line 36: I think we can leave this out for the moment until we launch the feature publicly.

content/v2/secondary-dns.markdown
```diff
@@ -0,0 +1,341 @@
...
+title: Secondary DNS API | DNSimple API v2
+excerpt: This page documents the DNSimple secondary DNS API v2.
+---
+
+# Secondary DNS API
+
+* TOC
+{:toc}
+
+
```
DXTimer, line 11:
```suggestion
<note>
<h4>This endpoint is currently in Public Preview. During the Preview period changes may occurr at any time.</h4>
</note>

```

layouts/sidebar.html
jacegu: Yes, @DXTimer is right. We will share the direct link with customers that we want to test the feature with, but it will not be announced to the general public yet.

content/v2/secondary-dns.markdown
jacegu: @weppos @aeden How do we want to say this? Should we treat the API as private beta?

aeden: Yes, you can replace "Public Preview" with "Private Beta". Also, "occur" is misspelled in that suggestion above. 😄

DXTimer: 😄 we should change the spelling at https://github.com/dnsimple/dnsimple-developer/blob/main/layouts/v2-preview.html

layouts/sidebar.html
duduribeiro: should we keep the `openapi` change so?

content/v2/secondary-dns.markdown
duduribeiro: I think we can drop `v2-preview`

layouts/sidebar.html
weppos: Yes, please to keep the OpenAPI. That's our primary source. It is reasonable to not link yet publish in the OpenAPI. The goal is not to hide, but simply to limit exposure until we're ready to advertise it.

DXTimer (approved):
Great work

sbastn (approved):

PR #331, opened 2021-03-25
Tabs component

sbastn:
This PR adds support for Vuejs components and introduces the first component: `<tabs>`. @DXTimer is planning to use this component in his use cases/tutorial series to display code in different languages. 

`<tabs>` supports any content and not only code.

## 👀 Preview

tabs

### 1. Write your content

Here is the basic template to get you started:

```html
<div class="component">
  <tabs>
    <template v-slot:ruby>Hello Ruby</template>
    <template v-slot:python>Hello Python</template>
  </tabs>
</div>
```

`tabs` supports markdown by default inside each `template` slot. If you're going to write an article with lots of code samples, I suggest that you declare the `.component` class at the top of your article, use `markdown="1"` for the content that is not inside the tabs. Like the following:

```html
<div class="component" markdown="1">
# First tutorial

Here's our hello world tutorial

  <tabs>
    <template v-slot:ruby>
~~~ruby
require 'dnsimple'
~~~
    </template>

    <template v-slot:python>
**Some Python code will be written!**
_(Eventually)_
    </template>
  </tabs>

# More explanations

  <tabs>
    <template v-slot:go>
*Go content*
    </template>
    <template v-slot:foo>
_foo content_
    </template>
  </tabs>

</div>

<script src="/dist/main.js"></script>
```

### 2. Render the component

Include `<script src="/dist/main.js"></script>` at the end of the page where `<tabs>` are used. Since this is the first draft of the component, I didn't add this script to the main layout. If the usage because more than a couple of pages, we should consider moving to the layout.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -12,7 +12,7 @@ BUILD_YARN_DIRECTORY = "dist"
 task default: [:test]
 
 desc "Compile the site"
-task compile: [:clean, :compile_nanoc, :compile_openapi]
+task compile: [:clean, :compile_nanoc, :compile_openapi, :compile_js]
 
 desc "Compile the static site"
 task :compile_nanoc do
@@ -38,6 +38,20 @@ task :compile_openapi do
   File.write("output/v2/openapi.json", JSON.dump(YAML.load(data)))
 end
 
+desc "Compile Javascript assets"
+task :compile_js do
+  puts "Compiling JS assets"
+
+  stdout, stderr, status = Bundler.with_unbundled_env do
+    Open3.capture3("./node_modules/.bin/webpack")
+  end
+  if status.success?
+    puts  "Compilation succeeded"
+  else
+    abort "ERROR: Compilation failed (#{$?.to_i}\n#{stdout}\n#{stderr}"
+  end
+end
+
 desc "Remove the compilation artifacts"
 task :clean do
   FileUtils.rm_r(PUBLISH_DIRECTORY) if File.exist?(PUBLISH_DIRECTORY)
@@ -52,7 +66,7 @@ end
 desc "Run the site"
 task run: [:compile] do
   Bundler.with_unbundled_env do
-    sh("bundle exec nanoc live")
+    sh("node node_modules/concurrently/bin/concurrently.js 'bundle exec nanoc live' './node_modules/.bin/webpack --watch'")
   end
 end
 
--- a/content/assets/css/_tabs.scss
+++ b/content/assets/css/_tabs.scss
@@ -1,0 +1,34 @@
+$border-color: #ccc;
+
+.tab-content {
+  border-left: 1px solid $border-color;
+  border-right: 1px solid $border-color;
+  border-bottom: 1px solid $border-color;
+  padding: 1rem;
+  pre {
+    padding: 0;
+    background: transparent;
+    border: 0;
+  }
+}
+
+.tabActive {
+  border-left: 1px solid $border-color;
+  border-right: 1px solid $border-color;
+  border-top: 1px solid $border-color;
+  border-bottom: none !important;
+}
+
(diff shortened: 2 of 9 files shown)


src/index.js
```diff
@@ -1 +1,13 @@
-// Empty for now. Needed for correct compilation with Webpack
+import Vue from "vue/dist/vue.esm.js"
+import Tabs from "./components/tabs.vue"
+
+// Troubleshoot with this
```
dallasread, line 4: Do we want to commit this?

dallasread (approved):
👍  Looks good, though I could not get the markdown to work (using the old Vue component, or my additions). I suspect this is a local configuration issue.

src/index.js
sbastn: I'll remove it. Thanks!

Rakefile
```diff
@@ -52,7 +66,7 @@ end
 desc "Run the site"
 task run: [:compile] do
   Bundler.with_unbundled_env do
-    sh("bundle exec nanoc live")
+    sh("node node_modules/concurrently/bin/concurrently.js 'bundle exec nanoc live' './node_modules/.bin/webpack --watch'")
```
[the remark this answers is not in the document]
sbastn: Yes, will do

Rakefile
```diff
@@ -38,6 +38,20 @@ task :compile_openapi do
   File.write("output/v2/openapi.json", JSON.dump(YAML.load(data)))
 end
 
+desc "Compile Javascript assets"
```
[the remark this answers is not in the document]
sbastn: It is

content/assets/css/_tabs.scss
```diff
@@ -0,0 +1,34 @@
+$border-color: #ccc;
```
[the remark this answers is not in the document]
sbastn: You're right. I can do better

sbastn:
After thinking about the feedback, I'm going to make this component nicer. I'm going to adapt @dallasread syntax and include `markdown` capability inside the component via https://marked.js.org/ so there is no need to use `markdown="1"`

sbastn:
@DXTimer @dallasread I reworked the `tabs` component to do the following:
1. Support markdown by default. You no longer need to explicitly declare markdown blocks inside each tab.
2. Removed the `tab0`, `tab1` structured naming to support better names.

Looking forward to your feedback!

src/components/tabs.vue
```diff
@@ -0,0 +1,39 @@
...
+    mounted () {
+      this.tabs = Object.entries(this.$slots).map(slot => {
+        return slot[0]; // returns the slot name, for instance, "ruby" for v-slot:ruby
+      });
+      const options = {
+        highlight: (code) => hljs.highlightAuto(code).value,
+      };
+      this.contents = Object.entries(this.$slots).map(slot => {
+        return marked(slot[1][0].text, options); // returns the content for the slot converted to markdown
+      });
```
dallasread, line 35: Would it be worth using `Object.keys` and `Object.values` (untested code)?

```suggestion
      this.tabs = Object.keys(this.$slots);
      const options = {
        highlight: (code) => hljs.highlightAuto(code).value,
      };
      this.contents = Object.values(this.$slots).map(text => {
        return marked(text, options); // returns the content for the slot converted to markdown
      });
```

dallasread (approved):
👍  LGTM!

One caveat with this approach is the single-word tab title limit, but I have no ideas on how to solve that at this time.

DXTimer (approved):
Great work! :shipit:

src/components/tabs.vue
sbastn: 👏 thanks

sbastn: 👏 thanks!

PR #334, opened 2021-04-05
Enhancement/open api spec update

DXTimer:


content/v2/openapi.yml
```diff
@@ -121,6 +122,13 @@ paths:
       description: Creates a domain and the corresponding zone into the account.
       parameters:
         - $ref: '#/components/parameters/Account'
+        - in: query
```
weppos, line 125: Beware this parameter is already defined in the `requestBody` section:

```
      requestBody:
        $ref: '#/components/requestBodies/DomainCreate'
```

PR #338, opened 2021-04-13
Mark getDomainPrices as Public preview

duduribeiro:
This commit marks our getDomainPrices endpoint as public preview.

image

### DIFF
--- a/content/v2/openapi.yml
+++ b/content/v2/openapi.yml
@@ -902,7 +902,7 @@ paths:
       summary: Get domain premium price
   '/{account}/registrar/domains/{domain}/prices':
     get:
-      description: Retrieves the domain price for registration, renewal, and transfer.
+      description: Retrieve domain prices.
       summary: Get prices for a domain
       parameters:
         - $ref: '#/components/parameters/Account'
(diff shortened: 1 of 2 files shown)


OleMchls:
I am ok with this change, but I am wondering what this _warning_ is about? @weppos ?  Screenshot 2021-04-13 at 23 20 49

OleMchls (approved):

weppos:
> I am ok with this change, but I am wondering what this _warning_ is about? @weppos ?

From what I can tell, that warning is likely wrong as this endpoint should completely replace the need to use the premium price endpoint (which we're likely going to deprecate). @DXTimer may know better as he's the one that published this initial docs.

weppos:
@duduribeiro can I suggest to short then title from:

> Retrieve domain price for registration, renewal, and transfer {#getDomainPrices}

to

> Retrieve domain prices {#getDomainPrices}

Same in the OpenAPI title. That will make it consistent with the current other endpoint titles. It's almost as if the title was also a description here.

Screenshot 2021-04-14 at 08 47 54

DXTimer:
> I am ok with this change, but I am wondering what this _warning_ is about? @weppos ? Screenshot 2021-04-13 at 23 20 49

This was added for consistency as with other endpoints that deal with pricing. I see now it's irrelevant, so it can be removed.

DXTimer (approved):
:shipit:

PR #347, opened 2021-05-20
[Secondary DNS] Remove the secondary zone delete endpoint

DXTimer:
Removes the secondary zone delete endpoint from the OpenAPI spec and the documentation.
In the documentation, a reference is left for the domain deletion endpoint.


sbastn (approved):
Thanks!

PR #354, opened 2021-07-02
Update documentation with endpoint rate limits

DXTimer:
Add missing documentation of checkDomain endpoint's rate limit, and make it known in the general section that such limits can apply.


weppos (changes requested):
I provided info to @DXTimer directly regarding why some info contained in this PR are potentially misleading.

DXTimer:
As discussed this PR is not relevant.

PR #365, opened 2021-09-09
Include public_key field in List and Show Delegation Signer Record API responses

lokst:
Include `public_key` field in sample responses for List and Show Delegation Signer Record API


ggalmazor (approved):
Looking good, thanks for updating these :)

OleMchls (approved):

PR #381, opened 2022-01-20
Adds the rust api client to the list of api clients

ecomba:
Adds a link to the rust api client in the Official clients section.

### DIFF
--- a/content/libraries.markdown
+++ b/content/libraries.markdown
@@ -23,6 +23,7 @@ _Please keep in mind that some of these libraries may be incomplete or oudated._
 - [DNSimple C# client](https://dnsimple.link/api-client-csharp)
 - [DNSimple PHP client](https://dnsimple.link/api-client-php) - development release
 - [DNSimple Python client](https://dnsimple.link/api-client-python) - development release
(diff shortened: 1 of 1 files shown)


weppos (approved):
👍 

FYI it looks like indentation is off by one, from the "DNSimple C# client" onwards.

Also, since we shipped and announced it, are we still in "development release"?
https://blog.dnsimple.com/2022/01/new-rust-api-client-library/
I suspect we're effectively in public beta phase. Can you confirm?

ecomba:
> 👍
> 
> FYI it looks like indentation is off by one, from the "DNSimple C# client" onwards.
> 
> Also, since we shipped and announced it, are we still in "development release"? https://blog.dnsimple.com/2022/01/new-rust-api-client-library/ I suspect we're effectively in public beta phase. Can you confirm?

I can't see the "off by one" in the code 🤷🏼

As of the rust client I would still leave it in "_development release_" as we are overhauling the error handling currently.

weppos:
> I can't see the "off by one" in the code 🤷🏼

Screenshot 2022-01-20 at 17 36 30

I think GitHub tricked me.

dallasread (approved):
Looks good to me!

PR #395, opened 2022-04-19
Bumps ruby to 3.0.4

sbastn:
I've added puma because Ruby 3.0 and later does not come with WEBrick by default so we can start the app locally.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
 # Update also: .travis.yml
-ruby 2.7.5
+ruby 3.0.4
 nodejs 16.13.0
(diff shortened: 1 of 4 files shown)


san983 (approved):

.tool-versions
```diff
@@ -1,3 +1,3 @@
 # Update also: .travis.yml
-ruby 2.7.5
+ruby 3.0.4
 nodejs 16.13.0
```
san983, line 3:
```suggestion
nodejs 16.14.2
```

Wanna take the opportunity to also update NodeJS to its LTS version?

PR #425, opened 2022-09-09
Improve documentation about error responses

jacegu:
This is an initial effort towards better documenting the errors our API returns. To do this we are:

1. Extending the existing section in the overview about errors. We are including specific error codes we use along examples of the different payloads to expect for each error type.
2. Creating an error sub-section for each endpoint in which we list the most relevant error codes the endpoint may return.
  - For now, we are leaving out generic errors like `404` and `429` that affect all endpoints. We are including 401, to make sure there is always an error section.
  - For each error type we are cross-linking the overview section about errors.

I have also made some edits to improve the consistency overall, including some wording and the section ordering.

### DIFF
--- a/content/v2.markdown
+++ b/content/v2.markdown
@@ -104,11 +104,11 @@ DNSimple uses conventional HTTP response codes to indicate success or failure of
 
 ### Error
 
-- `400 Bad Request` - A required parameter or the request is invalid.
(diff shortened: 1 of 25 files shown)


ggalmazor (approved):
💯

nestorsalceda (approved):
Thank you!

PR #426, opened 2022-09-13
Add OpenAPI response schema for 400 status code responses

DXTimer:
This PR documents the response schema when the API encounters a validation error.

### DIFF
--- a/content/v2/openapi.yml
+++ b/content/v2/openapi.yml
@@ -4780,6 +4780,17 @@ components:
           $ref: '#/components/headers/X-RateLimit-Remaining'
         X-RateLimit-Reset:
           $ref: '#/components/headers/X-RateLimit-Reset'
+      content:
+        application/json:
+          schema:
+            type: object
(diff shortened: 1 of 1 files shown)


jacegu:
@DXTimer I wonder if it's worth adding the same for other 4xx responses (mostly to replicate what I started doing [here](https://github.com/dnsimple/dnsimple-developer/pull/425/files#diff-a874e3881d21e9efbc86cb359304cbac1731b6847055384a4d03731f4df514a3))

DXTimer:
> @DXTimer I wonder if it's worth adding the same for other 4xx responses (mostly to replicate what I started doing [here](https://github.com/dnsimple/dnsimple-developer/pull/425/files#diff-a874e3881d21e9efbc86cb359304cbac1731b6847055384a4d03731f4df514a3))

I've updated the schema for the 401, 402, 404, and 429 status codes in the spec.

jacegu (approved):

PR #436, opened 2022-11-02
Document the Zone NS Records endpoint

dallasread:
This PR adds documentation for the new Zone NS Records endpoint.


content/v2/openapi.yml
```diff
@@ -1564,6 +1564,32 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
       summary: Check zone distribution
+  '/{account}/zones/{zone}/ns_records':
+    put:
+      description: Updates the zone name servers based on the provided name server set IDs.
```
dallasread, line 1569:
```suggestion
      description: Updates the zone's NS records
```

content/v2/openapi.yml
```diff
@@ -1564,6 +1564,32 @@ paths:
               schema:
                 $ref: '#/components/schemas/Error'
       summary: Check zone distribution
+  '/{account}/zones/{zone}/ns_records':
+    put:
+      description: Updates the zone's NS records
+      parameters:
+        - $ref: '#/components/parameters/Account'
+        - $ref: '#/components/parameters/Zone'
+        - $ref: '#/components/parameters/NsSetIds'
```
lokst, line 1573: Should `NsSetIds` be part of a `requestBody` instead?
How will `NsNames` be used, given that we have `NsSetIds` ?

content/v2/openapi.yml
```diff
@@ -1564,6 +1564,32 @@ paths:
...
+        - $ref: '#/components/parameters/NsNames'
```
lokst, line 1574: How will `NsNames` be used, given that we have `NsSetIds` ?

dallasread: The idea is that you can also provide ['ns1.example.com'] – like the form.

Forced to choose one, I tend to think the NsNames is more valuable to have because, if you're using the API, it is likely easier for you to manage the sets within your own source code. That being said, we'll also need to expand it for glue records (which wouldn't be verified as its not in our currently-laid-out work).

WDYT, remove the NsNames param?

lokst: I see, I think including both makes sense (for maximum flexibility), and I agree with just implementing support for  NsSetIds if we are constrained by time 🙂

kojdm (approved):
LGTM

lokst (approved):

OleMchls (approved):

san983 (approved):

content/v2/openapi.yml
```diff
@@ -4903,6 +4929,29 @@ components:
...
+              ns_names:
+                type: array
+                items:
+                  type: string
+                example:
+                  - ns1.dnsimple.com
+                  - ns2.dnsimple.com
+                  - ns3.dnsimple.com
+                  - ns4.dnsimple.com
+              ns_set_ids:
```
jacegu, line 4948: For clarification: Is it correct that NS Sets will only be created through the UI (at least for now)?

dallasread: Correct! There are no endpoints to manage name server sets.

We'll have to see how sets are used in the hands of customers, but my bet is that people will not have an API-requiring number of sets. Of course, now that I've said that, I'm sure that'll be the first feature request...

jacegu (approved):

DXTimer (approved):

content/v2/openapi.yml
```diff
@@ -4903,6 +4929,29 @@ components:
...
               - ns3.dnsimple.com
               - ns4.dnsimple.com
+    ZoneNsRecordsUpdate:
+      description: Zone name server attributes
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              ns_names:
```
weppos, line 4939: If you allow to pass sets, what's the purpose of allowing also individual name servers?

dallasread: The purpose is to be able to set a zone's NS records to values that aren't defined in a name server set. This way, you can add multiple name server sets and/or custom additional names. This matches the UI from the shape (and implementation thus far).

lokst (approved):

kojdm (approved):

PR #437, opened 2022-11-08
Promote record notes

sbastn:
I switched the Webauthn article for record notes and rotated the articles. The Netlify connector made out of the list.

The copy for the record notes came directly from a tweet and content on the blog article.

# Before
Screenshot 2022-11-08 at 10 32 51

# Now
Screenshot 2022-11-08 at 10 32 34


laetitiaeden (approved):
As per the wiki release process we are pushing the record notes announcement on the developer site

PR #472, opened 2023-03-09
Add missing DomainRegistration and DomainRenewal parameter types in OpenAPI spec

wilson-dnsimple:


DXTimer:
Nicely spotted. We should take this as an opportunity to add a linter to the spec.

DXTimer (approved):

PR #473, opened 2023-03-09
Lint OpenAPI Spec in the CI

DXTimer:
To avoid missing apparent OpenAPI spec violations, and to guide us towards a better spec this PR adds a lint step in the CI to inform a contributor if there are any OpenAPI spec violations.

We also ship a ignore file as we would need to allocate time to fix up the existing issues. With the exception of the problems that are being addressed by https://github.com/dnsimple/dnsimple-developer/pull/472. I will wait for those changes to be merged to `main` and then merge upstream to test workflow.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "devDependencies": {
     "@fortawesome/fontawesome-free": "^5.15.2",
+    "@redocly/cli": "^1.0.0-beta.123",
     "concurrently": "^6.0.0",
     "copy-webpack-plugin": "^11.0.0",
     "eslint": "^7.18.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -20,7 +21,8 @@
   },
   "scripts": {
     "live": "concurrently 'bundle exec nanoc live' 'bundle exec rake compile'",
-    "build": "./node_modules/.bin/webpack"
+    "build": "./node_modules/.bin/webpack",
```
wilson-dnsimple, line 24:
```suggestion
    "build": "npx webpack",
```

wilson-dnsimple (approved):

PR #484, opened 2023-04-20
Adds deprecation warning to whois privacy endpoints

juankuquintana:
This PR adds a deprecation warning to WhoisPrivacy endpoints.

1. [x] Get approval from Sebastian on UI
2. [x] Get approval from team on changes
3. [x] Get approval from Alyse on copy

### DIFF
--- a/content/v2/registrar/whois-privacy.markdown
+++ b/content/v2/registrar/whois-privacy.markdown
@@ -11,7 +11,7 @@ excerpt: This page documents the DNSimple WHOIS privacy API v2.
 Enable and disable WHOIS privacy on registered domains.
 
 <info>
-Now you can enable WHOIS Privacy protection for any of your domains any time, as long as the domain is registered with us and the TLD (Top Level Domain) supports the WHOIS Privacy service, the WHOIS Privacy service will stay enabled until it is disabled.
+Now you can enable WHOIS Privacy protection for any of your domains any time, as long as the domain is registered with us, and the TLD (Top Level Domain) supports WHOIS Privacy, the WHOIS Privacy service will stay enabled until it is disabled.
 </info>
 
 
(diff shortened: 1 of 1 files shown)


juankuquintana:
@sbastn It's me bothering you again 😬 . 

From a UI perspective, do you think this is enough to communicate our customers an endpoint is now deprecated?

content/v2/registrar/whois-privacy.markdown
```diff
@@ -17,6 +17,12 @@ Enable and disable WHOIS privacy on registered domains.
 
 Get the WHOIS privacy details for a domain.
 
+<warning>
+**The deprecation period for this endpoint began on 01 May 2023.**
+
+Now you can enable WHOIS Privacy protection for any of your domains any time, as long as the domain is registered with us and the TLD (Top Level Domain) supports the WHOIS Privacy service, the WHOIS Privacy service will stay enabled until it is disabled.
+</warning>
```
OleMchls, line 24: WDYT about pointing ppl to https://developer.dnsimple.com/v2/domains/#domain-attributes as that's where they would need to look if they want to understand if whois privacy is enabled

sbastn: @juankuquintana I would run this new content through @itsalyse. IMO it would be more valuable to announce the end date instead of the begin date of the deprecation.

I agree with @OleMchls suggestion about pointing to the end point that shows if the privacy is supported for their TLD.

sbastn (commented):
@juankuquintana run the new content through @itsalyse please. Thank you!

juankuquintana:
> @juankuquintana run the new content through @itsalyse please. Thank you!

Thanks @sbastn

OleMchls (approved):
🙌

kntsoriano (approved):

content/v2/registrar/whois-privacy.markdown
itsalyse: I think including both dates is a good idea - having the end date would definitely be helpful here

itsalyse:
I left one comment but otherwise the copy lgtm 👍

PR #517, opened 2023-08-10
Document optional "integrated_zones" parameter and behavior in Zone Records API

lokst:
Document the optional `integrated_zones` parameter and behavior in Zone Records API

### DIFF
--- a/content/v2/openapi.yml
+++ b/content/v2/openapi.yml
@@ -1822,6 +1822,8 @@ paths:
       operationId: deleteZoneRecord
       tags:
         - zones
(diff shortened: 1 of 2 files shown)


dallasread (approved):
LGTM – this is awesome.

Could we add the `integrated_zones` param in the create / update examples?

lokst:
> Could we add the integrated_zones param in the create / update examples?

@dallasread Great idea, just added 👍

kojdm (approved):
This LGTM! Thanks 🎉

PR #521, opened 2023-08-22
Fixtures for contact management API

DXTimer:
Add example responses for the Registrant Change API.

Belongs to https://github.com/dnsimple/dnsimple-business/issues/1729

### DIFF
--- a/fixtures/v2/api/deleteRegistrantChange/success.http
+++ b/fixtures/v2/api/deleteRegistrantChange/success.http
@@ -1,0 +1,13 @@
+HTTP/2 201 
+server: nginx
+date: Tue, 22 Aug 2023 11:14:44 GMT
+content-type: application/json; charset=utf-8
+x-ratelimit-limit: 2400
+x-ratelimit-remaining: 2391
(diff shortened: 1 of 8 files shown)


fixtures/v2/api/deleteRegistrantChange/success.http
```diff
@@ -0,0 +1,13 @@
+HTTP/2 201 
+server: nginx
+date: Tue, 22 Aug 2023 11:14:44 GMT
+content-type: application/json; charset=utf-8
```
wilson-dnsimple, line 4: This doesn't seem correct, as the body is not JSON.

DXTimer: That is the correct response we receive from the backend. The client should be able to handle these responses. However, I do agree that something is not right about the response, since the correct response would be 204.

wilson-dnsimple (approved):
We must remember to regenerate once https://github.com/dnsimple/dnsimple-app/issues/23629 is fixed.

PR #523, opened 2023-08-29
Remove redundant tags in OpenAPI spec

wilson-dnsimple:
A few endpoints have two tags, one of which is redundant. This is not consistent with all other endpoints, which only have one specific tag.

This also fixes an outlier usage of `nullable` instead of the existing aliased type.

Belongs to https://github.com/dnsimple/dnsimple-business/issues/1728


DXTimer (approved):
I can see that we inconsistently apply the `registrar` tag to operations that fall under the category, so I agree with cleaning it up.

PR #524, opened 2023-08-30
Update deleteRegistrantChange responses and fixtures

DXTimer:
This PR updates the OpenAPI definition for the **deleteRegistrantChange** operation. It changes the responses to:

Before:

- Return 200 with an empty body when a cancellation happens sync
- Return 201 with an empty body when a cancellation is happening async

Now:

- Return 204 with an empty body when a cancellation happens sync
- Return 202 with the registrant change record when a cancellation happens async

Fixtures have been updated to reflect the changes.

Belongs to https://github.com/dnsimple/dnsimple-business/issues/1729
Depends on https://github.com/dnsimple/dnsimple-app/pull/23686

### DIFF
--- a/fixtures/v2/api/deleteRegistrantChange/success.http
+++ b/fixtures/v2/api/deleteRegistrantChange/success.http
@@ -1,4 +1,4 @@
-HTTP/1.1 201
+HTTP/1.1 204 No Content
 server: nginx
 date: Tue, 22 Aug 2023 11:14:44 GMT
 content-type: application/json; charset=utf-8
--- a/content/v2/openapi.yml
+++ b/content/v2/openapi.yml
@@ -2123,10 +2123,17 @@ paths:
       tags:
         - registrant changes
       responses:
-        '200':
+        '204':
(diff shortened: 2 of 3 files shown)


fixtures/v2/api/deleteRegistrantChange/success.http
```diff
@@ -1,4 +1,4 @@
-HTTP/1.1 201
+HTTP/1.1 204 No Content
 server: nginx
 date: Tue, 22 Aug 2023 11:14:44 GMT
 content-type: application/json; charset=utf-8
```
wilson-dnsimple, line 4: Ideally, this should still be omitted, as the content is not JSON (an empty string is not valid JSON).

wilson-dnsimple (approved):

fixtures/v2/api/deleteRegistrantChange/success.http
DXTimer: Agree, would you mind opening a separate issue in the API backend repo to track this change we want to make?

PR #530, opened 2023-09-26
Add linked provider param to domain registration API

kntsoriano:
This PR adds the `linked_provider` parameter to the domain registration endpoint. If a `linked_provider` param is passed, we register the domain through the given provider. Otherwise, the domain is registered through DNSimple.

As we didn't have time to completely consult the external registrations team as to how the API is designed, I opened this PR to facilitate that discussion. Should we change the design?

FYI, the API endpoint for a domain transfer via a linked provider has not been implemented yet.

### DIFF
--- a/content/v2/openapi.yml
+++ b/content/v2/openapi.yml
@@ -3608,8 +3608,6 @@ components:
           type: boolean
         whois_privacy:
           type: boolean
-        linked_provider:
-          type: string
         created_at:
           $ref: '#/components/schemas/DateTime'
         updated_at:
@@ -5573,6 +5571,9 @@ components:
               premium_price:
                 type: string
                 description: 'Required as confirmation of the price, only if the domain is premium.'
+              linked_provider:
+                type: string
+                description: 'The nickname of a linked provider to register the domain through. If left blank, registration will go through DNSimple.'
             example:
               registrant_id: 1
     DomainRenew:
--- a/content/v2/openapi.yml
+++ b/content/v2/openapi.yml
@@ -3608,6 +3608,8 @@ components:
           type: boolean
         whois_privacy:
           type: boolean
+        linked_provider:
+          type: string
         created_at:
           $ref: '#/components/schemas/DateTime'
         updated_at:
--- a/content/v2/registrar.markdown
+++ b/content/v2/registrar.markdown
@@ -198,6 +198,7 @@ Name | Type | Description
 -----|------|------------
 `registrant_id` | `integer` | **Required**. The ID of an existing [contact](/v2/contacts/#contact-attributes) in your account.
 `whois_privacy` | `bool` | Set to true will attempt to purchase/enable the whois privacy as part of the registration. An extra cost may apply. Default: `false`.
+`linked_provider` | `string` | When set to the nickname of a linked provider, registration of the domain will go through that provider. Otherwise, registration of the domain will go through DNSimple.
 `auto_renew` | `bool` | Set to true to enable the auto-renewal of the domain. Default: `false`.
 `extended_attributes` | `hash` | **Required** for TLDs that require [extended attributes](/v2/tlds/#getTldExtendedAttributes).
 `premium_price` | `string` | **Required** as confirmation of the price, only if the domain is premium.


content/v2/openapi.yml
```diff
@@ -3608,6 +3608,8 @@ components:
           type: boolean
         whois_privacy:
           type: boolean
+        linked_provider:
+          type: string
```
weppos, line 3612: I recommend we set the same description also here. I believe the field should have some sort of summary or description field.

content/v2/registrar.markdown
```diff
@@ -198,6 +198,7 @@ Name | Type | Description
 -----|------|------------
 `registrant_id` | `integer` | **Required**. The ID of an existing [contact](/v2/contacts/#contact-attributes) in your account.
 `whois_privacy` | `bool` | Set to true will attempt to purchase/enable the whois privacy as part of the registration. An extra cost may apply. Default: `false`.
+`linked_provider` | `string` | When set to the nickname of a linked provider, registration of the domain will go through that provider. Otherwise, registration of the domain will go through DNSimple.
```
weppos, line 201: How are provider created? Do you have a screenshot? Is the field called "nickname"? We use more often terms like label, name, or else.

AGS4NO: The field on creation asks for a Nickname. Once created the nickname is shown under the Linked Providers section:

Screenshot 2023-09-28 at 11 34 41 AM

- I think `provider_nickname` might be more explicit for the name
- Is it possible to use the same nickname for different providers? What If I want a `default` nickname for each linked provider?

From a user perspective, If I have multiple providers each with multiple accounts linked, I am heavily relying on nicknames to determine which provider/account combination I am using. 

From an operational/readability perspective:

```
provider : "route53"
provider_nickname : "default"
```

OR

```
linked_provider: "route53:default"
```

OleMchls: It’s rather uncommon indeed. I like it because it makes the distinction to name clear. As the provider name is GoDaddy and we let customers assign an account-unique “nickname”. I think another maybe more “common” name would be alias.

content/v2/openapi.yml
```diff
@@ -3608,6 +3608,8 @@ components:
           type: boolean
         whois_privacy:
           type: boolean
+        linked_provider:
```
DXTimer, line 3611: We should also update the example. This is the Domain Registration model and not the RegisterDomain endpoint. If I understand correctly it also needs to be updated with the new parameter.

content/v2/registrar.markdown
DXTimer: Currently, nicknames are unique within an account, so I do not see an issue using them as an identifier.

kntsoriano: What should we use for the param name? I think `provider_nickname` might be more clear and is developer-friendly with the conventions we are used to.

content/v2/openapi.yml
kntsoriano: Thanks! I will move this as this was in the wrong place.

content/v2/openapi.yml
kntsoriano: Thanks! I fixed this now.

DXTimer: @kntsoriano does this mean that we do not return the provider this domain is registered within our domain model? If so have you considered adding it to the model?

My thinking on this is that as an integration developer, I can use the provider nickname to at the very least create a local mapping of Nickname -> GoDaddy, DNsimple, etc. This information could be used to visualize the domains differently or group them in my UI or as a filter for certain operations or policies.

kntsoriano: At the moment, we don't return it. I think it's a good idea to return the provider nickname inside the domain model :+1:

content/v2/registrar.markdown
DXTimer: Are you considering adding additional validation on the characters allowed to be used in the nickname?

kntsoriano: We have not considered adding any validations except the ones that we have right now. The only validation we have right now is that we require the nickname to be present.

kntsoriano:
I will be merging this now. We will update the code to reflect the suggestions and then update the docs again after.

PR #536, opened 2023-10-20
Billing Charges API

DXTimer:
This PR adds OpenAPI definitions for the new Billing Charges API.

There is a single endpoint that has been added:

1. `GET /:account/billing/charges` - (`listCharges`) Lists the billing charges for the account.

Belongs to https://github.com/dnsimple/dnsimple-business/issues/1763

### DIFF
--- a/content/v2/billing-charges.markdown
+++ b/content/v2/billing-charges.markdown
@@ -76,7 +76,7 @@ curl  -H 'Authorization: Bearer <token>' \
 Responds with HTTP 200.
 
 ~~~json
-<%= pretty_print_fixture("/api/listCharges/success.http") %>
+pretty_print_fixture("/api/listCharges/success.http")
 ~~~
 
 ### Errors
--- a/content/v2/billing-charges.markdown
+++ b/content/v2/billing-charges.markdown
@@ -30,10 +30,10 @@ GET /:account/billing/charges
 
 ### Filters
 
-| Name          | Description                                 |
-| ------------- | ------------------------------------------- |
-| `:start_date` | Only include results after the given date.  |
-| `:end_date`   | Only include results before the given date. |
+| Name          | Description                                                                 |
+| ------------- | --------------------------------------------------------------------------- |
+| `:start_date` | Only include results after the given date. Format is ISO8601 (YYYY-MM-DD).  |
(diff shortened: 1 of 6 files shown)


AGS4NO (approved):

fixtures/v2/api/listCharges/success.http
```diff
@@ -0,0 +1,14 @@
...
+Connection: close
+X-RateLimit-Limit: 2400
+X-RateLimit-Remaining: 2397
+X-RateLimit-Reset: 1698143967
+Content-Type: application/json; charset=utf-8
+X-WORK-WITH-US: Love automation? So do we! https://dnsimple.com/jobs
+Cache-Control: no-store, must-revalidate, private, max-age=0
+X-Request-Id: a57a87c8-626a-4361-9fb8-b55ca9be8e5d
+X-Runtime: 0.060526
+Transfer-Encoding: chunked
```
wilson-dnsimple, line 12: Is this correct? The body does not appear to match the [chunked encoding format](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#chunked). This appears to break the Java tests. Apples to the other fixtures too.

DXTimer: Fixtures have been updated. And indeed we do not use chunked encoding in our live environments.

content/v2/billing-charges.markdown
```diff
@@ -0,0 +1,88 @@
...
+
+| Name       | Type      | Description    |
+| ---------- | --------- | -------------- |
+| `:account` | `integer` | The account ID |
+
+### Filters
+
+| Name          | Description                                 |
+| ------------- | ------------------------------------------- |
+| `:start_date` | Only include results after the given date.  |
```
weppos, line 35: We should provide an example of the format.

weppos: Actually it's in the examples below, so it may be good. Although I would still add a format info.

weppos (approved):

content/v2/billing-charges.markdown
DXTimer: Format information added in https://github.com/dnsimple/dnsimple-developer/pull/536/commits/1234b288936a4c74816ccfb65c360e7e65a3b32f

PR #558, opened 2024-02-01
Add new DNS analytics endpoint to OpenAPI spec 

jacegu:
Closes dnsimple/dnsimple-app#25676

This PR:
- Updates the OpenAPI spec with the new endpoint.

> [!NOTE]  
> This is my first time updating the OpenAPI spec so expect rough edges.
> 
> Note that I chose not to extract much to be referenced and write the spec pretty much in line: I do not think much of what we have in this endpoint is currently reusable.

### DIFF
--- a/content/v2/openapi.yml
+++ b/content/v2/openapi.yml
@@ -3101,7 +3101,6 @@ components:
           - 'zone_name:asc'
           - 'zone_name:desc'
     SortDomains:
-    SortDomains:
       description: Sort results. Default sorting is ascending by name.
       name: sort
       style: form
--- a/content/v2/openapi.yml
+++ b/content/v2/openapi.yml
@@ -2756,8 +2756,10 @@ paths:
                       rows:
                         type: array
                         items:
-                          type:
-                            - string | integer
+                          anyOf:
+                            - type: integer
+                            - type: string
+                            - $ref: '#/components/schemas/Date'
                           example: [example.com,102985]
                       headers:
                         type: array
--- a/content/v2/openapi.yml
+++ b/content/v2/openapi.yml
@@ -2699,7 +2699,9 @@ paths:
       summary: Delete a webhook
   '/{account}/dns_analytics':
     get:
-      description: Retrieve data about DNS queries for the zones in an account.
+      description:
+        Queries and returns DNS Analytics data available for the provided query parameters.
+        This API is currently in Public Beta. During the Public Beta period changes may occur at any time.
       parameters:
         - $ref: '#/components/parameters/Account'
         - startDate:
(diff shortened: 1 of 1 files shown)


content/v2/openapi.yml
```diff
@@ -2697,6 +2697,107 @@ paths:
...
+                    type: object
+                    properties:
+                      rows:
+                        type: array
+                        items:
+                          anyOf:
+                            - type: integer
+                            - type: string
+                            - $ref: '#/components/schemas/Date'
+                          example: [example.com,102985]
```
ggalmazor, line 2763:
```suggestion
                          example: ["example.com", "2024-08-01", 102985]
```
A couple of things:
- Add a date for a complete example
- Use double quote wrapper for strings for JSON compliance

content/v2/openapi.yml
```diff
@@ -2697,6 +2697,107 @@ paths:
...
+                      headers:
+                        type: array
+                        items:
+                          type: string
+                          example: zone_name,volume
```
ggalmazor, line 2768:
```suggestion
                          example: "zone_name,volume"
```

content/v2/openapi.yml
```diff
@@ -2963,6 +3083,21 @@ components:
...
+      required: false
+      schema:
+        type: string
+        enum:
+          - 'date:asc'
+          - 'date:desc'
+          - 'volume:asc'
+          - 'volume:desc'
+          - 'zone_name:asc'
+          - 'zone_name:desc'
```
ggalmazor, line 3100: Multiple sorting criteria can be defined. In fact, the default is `"date:asc,zone_name:asc"`. Can we just not provide the `enum` here?

ggalmazor (changes requested):
LGTM overall. 

I left some suggestions with a few things I think we need to tweak

content/v2/openapi.yml
jacegu: I just followed the example of other endpoints in which we are providing the enum, even when the values can be combined. Not sure if I did the right thing.

ggalmazor: I'm not sure how to interpret `enum` either...

ggalmazor: I _think_ that we're not using them correctly if values can be combined https://swagger.io/docs/specification/data-models/enums/

It sounds like we would need to list all permutations under all possible sorting, which feels wrong to do. Maybe we should not have `enum` unless there's a finite list of values our customers need to pick from.

jacegu: I was looking [at the same thing](https://swagger.io/docs/specification/data-models/enums/) and I agree: I do not think this is the correct use. Will update this momentarily.

jacegu: See: [1de58d1](https://github.com/dnsimple/dnsimple-developer/pull/558/commits/1de58d1563f6ca6a2cbc214d0bdf9aaa63f7b038)

ggalmazor (approved):
LGTM!

juankuquintana (approved):
👍 You could run it into https://editor.swagger.io/ to validate.

I did, and apparently we have an issue with another endpoint but nothing was raised with the new changes.

itsalyse (approved):

PR #582, opened 2024-03-08
Fix curl examples that were missing the account ID

ggalmazor:
In this PR:
- Fix some curl examples that were missing the account ID

### DIFF
--- a/content/v2/dns-analytics.markdown
+++ b/content/v2/dns-analytics.markdown
@@ -80,7 +80,7 @@ Pagination is supported in this endpoint:
 ~~~
 curl  -H 'Authorization: Bearer <token>' \
       -H 'Accept: application/json' \
-      https://api.dnsimple.com/v2/dns_analytics
+      https://api.dnsimple.com/v2/1385/dns_analytics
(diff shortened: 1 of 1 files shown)


content/v2/dns-analytics.markdown
```diff
@@ -80,23 +80,23 @@ Pagination is supported in this endpoint:
 ~~~
 curl  -H 'Authorization: Bearer <token>' \
       -H 'Accept: application/json' \
-      https://api.dnsimple.com/v2/dns_analytics
+      https://api.dnsimple.com/v2/1385/dns_analytics
```
jacegu, line 83:
```suggestion
      https://api.dnsimple.com/v2/<account_id>/dns_analytics
```
☝️ shouldn't we parametrize this like we do with `token`?

ggalmazor: I've put the same thing we're using in other dev docs...

jacegu: then I'll 🤐

jacegu (approved):

PR #762, opened 2025-02-18
Improving Search Experience: Use the search widget

kojdm:
Clicking the search bar now opens the support widget. Also removes old search.js.

https://deploy-preview-762--dnsimple-developer.netlify.app/


layouts/header.html
```diff
@@ -20,7 +20,7 @@
...
             </a>
-            <input type="search" name="q" value="" id="input-search" class="input-search input-reset outline-0 w-100 w-50-l br-pill pv2 ph3 mt3 mt0-ns ml2-ns" placeholder="Search..." autocomplete="off">
+            <input type="search" name="q" value="" id="input-search" class="input-search input-reset outline-0 w-100 ba br-pill b--black-20 mb0 pv2 ph3 ml2-ns" placeholder="Search our developer docs (⌘K)" autocomplete="off" data-dnsimple-open-support-widget>
```
dallasread, line 23: Is it possible to use the same approach as the support site? https://github.com/dnsimple/dnsimple-support/commit/bcbe9290635cf927b3b384ebd1fa2601003d5e96#diff-6ab499c3b9febcf6798441d6e51489f18bb2a82f099e2e1435a898943a835a7cL33

dallasread (approved):
Works well! 👏

layouts/header.html
dallasread: May be even better (accessiblity-wise) to use a <button> for both.

PR #803, opened 2025-05-22
Shared Domains deprecation

dallasread:
Since we've had this feature deprecated for some time, we're removing mentions of it in the app. It seems to make sense to follow suit on the dev site as well.

Belongs to https://github.com/dnsimple/dnsimple-app/issues/27411


_redirects
```diff
@@ -1,2 +1,4 @@
 /v2/webhooks /v2/webhooks/webhooks
 /v2/webhooks/index.html /v2/webhooks/webhooks/
+/v2/domains/collaborators /v2/domains/
+/v2/domains/collaborators/index.html /v2/domains/
```
jacegu, line 4: Had no idea we were doing this 😮

dallasread: I started off just leaving a small hint of deprecation on the collaborators page. Then, I said this is silly and found this redirects file!

PR #808, opened 2025-06-02
adding test that enforces trailing slashes to internal urls

dperrymorrow:
adding tests to ensure that we have trailing slashes on our local URLs to prevent redirects hurting SEO


sbastn (approved):
Yep!

PR #813, opened 2025-06-12
Adding marketing components to developer site

dperrymorrow:
Now that the new marketing components are sharable from dnsimple.com, adding them to the blog so that they are in sync with what is done in the admin.

belongs to: https://github.com/dnsimple/dnsimple-marketing/issues/901


dperrymorrow:
Screenshot 2025-06-12 at 2 14 00 PM

PR #816, opened 2025-06-13
fix: Update broken fixtures

weppos:
Over the last few years we made several changes to fixtures, sometimes even "manual", and we never ported them to clients.

When I tried to pull the fixtures from here into the dnsimple-go client... [it was a bloodbath](https://github.com/dnsimple/dnsimple-go/pull/200).

I fixed all fixtures in the Go client repo, and I am porting the fixtures here.

> [!IMPORTANT]
> This repository is the source of truth for fixures. Fixtures MUST be updated here, then MUST be pushed down to all clients that will be adjusted to work on the new fixtures.
>
> Clients should not make changes to fixtures directly.


lokst (approved):

kojdm (approved):
Thanks for this!

PR #843, opened 2025-08-14
Fix content-length values in webhook fixtures

lokst:
Currently the tests in the dnsimple-go rely on the Content-Length values being accurate.
I have fixed the values by using the actual content length:

```
python3 -c "import re; p='webhooks/email_forward.deactivate/example.http'; b=open(p,'rb').read(); m=re.search(rb'\r?\n\r?\n', b); print(len(b[m.end():]) if m else 0)"
```


lokst:
@kntsoriano Going ahead to merge this as it's a tiny change - let me know if you spot anything amiss!

PR #848, opened 2025-08-21
Update dnssec rotation events

weppos:
DNSSEC events include a zone since June. The create/delete events were updated as part of https://github.com/dnsimple/dnsimple-developer/commit/0fc2ea82d9a76f437bec1216f1a2bed67dd3ff1b, whereas the start/complete rotation events were not.

Once merged, will need to expand https://github.com/dnsimple/dnsimple-go/blob/012db326c1287dca484f8e7b878df04be97e9d6b/dnsimple/webhook/events.go#L158-L166

Blocks https://github.com/dnsimple/strillone/issues/65


lokst (approved):

kntsoriano (approved):

kntsoriano:
@lokst Maybe a further QA step for here would be to attempt to use the fixture in one of the tests in one of our API clients? Just to ensure that the fixtures are valid.

kntsoriano:
@lokst That's a better idea 👍

PR #863, opened 2025-09-22
Document batch change zone records endpoint

lokst:


content/v2/zones/records.markdown
```diff
@@ -309,6 +309,150 @@ Responds with [HTTP 400](/v2/#bad-request) if the record cannot be deleted.
...
+      -H 'Accept: application/json' \
+      -H 'Content-Type: application/json' \
+      -X POST \
+      -d '<json>' \
+      https://api.dnsimple.com/v2/1010/zones/example.com/batch
+~~~
+
+### Input
+
+All operation types are optional. You can specify any combination of `creates`, `updates`, and `deletes`.
```
jacegu, line 347:
```suggestion
All operation types are optional. You can specify any combination of `creates`, `updates`, and `deletes`, as long as you respect the maximum batch size.
```

content/v2/zones/records.markdown
```diff
@@ -309,6 +309,150 @@ Responds with [HTTP 400](/v2/#bad-request) if the record cannot be deleted.
 
 Responds with [HTTP 401](/v2/#unauthorized) in case of case of authentication issues.
 
+## Batch change zone records {#batchChangeZoneRecords}
+
+~~~
+POST /:account/zones/:zone/batch
+~~~
+
+This endpoint allows you to perform multiple record operations (create, update and/or delete) for a zone in a single request for efficiency and atomicity.
```
jacegu, line 318: We probably should mention that there is a limit to the number of operations you can perform in a single batch, and that you can combine them in any way you want.

content/v2/zones/records.markdown
```diff
@@ -309,6 +309,150 @@ Responds with [HTTP 400](/v2/#bad-request) if the record cannot be deleted.
...
+
+<info>
+The batch change zone records API is a feature that is only available to the Enterprise plan.
+If the feature is not enabled, you will receive an HTTP 412 response code.
+</info>
```
jacegu, line 323: 👍

content/v2/zones/records.markdown
jacegu: ⚠️ Another *very important* thing to mention here, is that customers won't be able to use this endpoint with zones that have integrated providers.

PR #882, opened 2025-10-22
Restore Google Analytics 4

sbastn:
Developer site doesn't use Google Tag Manager like dnsimple-app. We use the GA4 property directly.

Belongs to https://github.com/dnsimple/dnsimple-marketing/issues/983

## QA

1. Open the Netlify preview link
2. Verify that the GA4 tag appears in the `<head>`
3. The GA4 tag should not appear when running the app locally

### DIFF
--- a/layouts/analytics.html
+++ b/layouts/analytics.html
@@ -12,6 +12,17 @@
 </script>
 <% end %>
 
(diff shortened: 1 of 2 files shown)


dperrymorrow (approved):
verified

PR #909, opened 2026-01-07
Remove unused `eslint`

weppos:
We have 2 active security alerts caused by:

https://github.com/dnsimple/dnsimple-developer/security/dependabot/94
https://github.com/dnsimple/dnsimple-developer/security/dependabot/93

It looks like this is a transitive dependency of `eslint`:

> eslint 7.32.0  js-yaml 3.14.1

We use a very old eslint version, it's behind by 2 major versions. However, while searching this repo for eslint usage, I could not find any usage. Hence I'm removing it, unless we find it useful and we need to incorporate it in the workflow. At that point we may want to start fresh with a more recent version.

Note the third alert is also caused by eslint: https://github.com/dnsimple/dnsimple-developer/security/dependabot/96


dallasread (approved):
As this repo has JS, it would be nice to have the linter. Agree that there is no point having it there if it is not hooked up to anything.

PR #918, opened 2026-01-13
Add fixture for subscription.renew

weppos:
In https://github.com/dnsimple/dnsimple-go/pull/235 I added support to the remaining events defined by our app, according to our fixtures. What I found, is that `subscription.renew` did not have a proper fixture, just an "error" one.

This PR adds an example fixture as common. 

I actually looked at the other fixture as well, and I can't really determine what the "state failed" means. It seems a successful state. Perhaps we can remove it?

> [!NOTE]
> At some point we need to compare https://github.com/dnsimple/dnsimple-developer/blob/main/content/v2/openapi.yml#L5931 with the list of fixtures we publish, as well with list of events we actually emit from the app.

### DIFF
--- a/fixtures/v2/webhooks/subscription.renew/example.http
+++ b/fixtures/v2/webhooks/subscription.renew/example.http
@@ -1,0 +1,13 @@
+POST /1djlwbe1 HTTP/1.1
+Host: example.com
+Content-Type: application/json
+Total-Route-Time: 7210
+Content-Length: 432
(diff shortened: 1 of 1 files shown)


weppos:
@ggalmazor any idea about:

> I actually looked at the other fixture as well, and I can't really determine what the "state failed" means. It seems a successful state. Perhaps we can remove it?

PR #1031, opened 2026-06-13
Document public-client (PKCE) OAuth flow and CLI browser login

weppos:
DNSimple introduced a [new OAuth mechanism for the CLI](https://github.com/dnsimple/cli/issues/42), along with changes to the OAuth flow itself. This updates the customer-facing developer docs to match.

On the server, the OAuth provider now supports **public clients** that authenticate with **PKCE** (RFC 7636, `S256`) instead of a `client_secret`, plus **loopback redirect URIs** with dynamic ports (RFC 8252 §7.3) for native and CLI apps. On the CLI side, `dnsimple auth login` now **defaults to an interactive browser login**, with API token login available via `--with-token`.

Related to https://github.com/dnsimple/dnsimple-support/pull/1984

### DIFF
--- a/content/v2/oauth.md
+++ b/content/v2/oauth.md
@@ -18,7 +18,7 @@ Developers need to register their application before getting started. A register
 - **Public clients** are native, mobile, single-page, or command-line apps that cannot keep a secret. They authenticate with [PKCE](https://www.rfc-editor.org/rfc/rfc7636) (Proof Key for Code Exchange) instead of a client secret. See [Native and CLI apps](#native-and-cli-apps-loopback-redirect-uris).
 
 > [!NOTE]
-> Creating a public OAuth application is currently available upon request. Contact DNSimple support to enable public clients for your account.
+> If you do not see a client type choice when you register an application, your account registers web apps (confidential clients) by default. Public clients can be enabled on request; contact DNSimple support to register one for your account.
 
 ## Web Application Flow
 
@@ -117,9 +117,6 @@ $ curl -H "Authorization: Bearer ACCESS-TOKEN" https://api.dnsimple.com/v2/whoam
 ## Native and CLI apps (loopback redirect URIs)
 
 Native applications such as command-line tools and desktop apps are public clients: they cannot keep a client secret, so they secure the flow with [PKCE](https://www.rfc-editor.org/rfc/rfc7636) and receive the authorization `code` on a loopback callback running on the local machine.
-
-> [!NOTE]
-> Creating a public OAuth application is currently available upon request. Contact DNSimple support to enable public clients for your account.
 
 Register a loopback callback URL as your application's redirect URI:
 
--- a/content/v2/oauth.md
+++ b/content/v2/oauth.md
@@ -12,7 +12,13 @@ excerpt: This page documents the OAuth 2 flow you can use to access the DNSimple
 
 DNSimple supports [OAuth 2.0](http://oauth.net/) for authenticating API requests. OAuth 2 is a protocol that lets external applications request authorization to private details in a user's DNSimple account without getting their password. This is preferred over Basic Authentication because tokens can be revoked by users at any time.
 
-Developers need to register their application before getting started. A registered OAuth application is assigned a unique client ID and client secret. The client secret should not be shared.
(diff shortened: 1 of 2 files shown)


content/v2/oauth.md
```diff
@@ -102,3 +113,24 @@ $ curl -H "Authorization: Bearer ACCESS-TOKEN" https://api.dnsimple.com/v2/whoam
 
 > [!NOTE]
 > If you are using the [sandbox environment](/sandbox/) replace `dnsimple.com` with `sandbox.dnsimple.com`.
+
+## Native and CLI apps (loopback redirect URIs)
+
+Native applications such as command-line tools and desktop apps are public clients: they cannot keep a client secret, so they secure the flow with [PKCE](https://www.rfc-editor.org/rfc/rfc7636) and receive the authorization `code` on a loopback callback running on the local machine.
+
+> [!NOTE]
+> Creating a public OAuth application is currently available upon request. Contact DNSimple support to enable public clients for your account.
```
Pemacaa: This note appears twice in the same article, and I think that we can get rid of this one.

content/v2/oauth.md
```diff
@@ -12,14 +12,22 @@ excerpt: This page documents the OAuth 2 flow you can use to access the DNSimple
...
 
-Developers need to register their application before getting started. A registered OAuth application is assigned a unique client ID and client secret. The client secret should not be shared.
+Developers need to register their application before getting started. A registered OAuth application is assigned a unique client ID, and the way it authenticates depends on its type:
+
+- **Confidential clients** are server-side web apps that can keep a secret. They are assigned a client ID and a client secret, and authenticate at the token endpoint with that secret. The client secret should not be shared.
+- **Public clients** are native, mobile, single-page, or command-line apps that cannot keep a secret. They authenticate with [PKCE](https://www.rfc-editor.org/rfc/rfc7636) (Proof Key for Code Exchange) instead of a client secret. See [Native and CLI apps](#native-and-cli-apps-loopback-redirect-uris).
+
+> [!NOTE]
+> Creating a public OAuth application is currently available upon request. Contact DNSimple support to enable public clients for your account.
```
Pemacaa: For consistency with the support article (#1984): support says "if you do not see a client type choice, your account registers web apps by default," while this says "available upon request, contact support." Same idea, but slightly different wording. Might be worth aligning so customers see the same message in both places.

content/v2/oauth.md
```diff
@@ -102,3 +113,24 @@ $ curl -H "Authorization: Bearer ACCESS-TOKEN" https://api.dnsimple.com/v2/whoam
...
+
+```
+http://127.0.0.1/callback
+http://[::1]/callback
+```
+
+A native app binds its callback server to an ephemeral port chosen at runtime, so it cannot register that port in advance. As described in [RFC 8252 section 7.3](https://www.rfc-editor.org/rfc/rfc8252#section-7.3), DNSimple ignores the port of a loopback redirect URI when both the registered URI and the runtime URI use a loopback IP literal. Your app can therefore listen on any available port, for example `http://127.0.0.1:53241/callback`, and the redirect URI will still match.
+
+> [!NOTE]
+> Port leniency applies only to the loopback IP literals `127.0.0.1` and `[::1]`. The hostname `localhost` is matched strictly, including its port, so use an IP literal if you need a dynamic port.
```
Pemacaa, line 131: I checked in the app when creating an OAuth application, and the help text there says : "You can use HTTP with localhost, or with a loopback address (127.0.0.1, [::1]) for native apps like CLIs; the registered port is matched leniently in that case."

The support article in dnsimple/dnsimple-support#1984 matches that behavior. This PR says localhost is matched strictly, including its port. Should the developer docs be aligned with the app UI and the support article?

Pemacaa (approved):

content/v2/oauth.md
weppos: Good catch. I think actually support and UI are wrong. Let me check.

content/v2/oauth.md
weppos: Fixed

content/v2/oauth.md
weppos: Fixed

content/v2/oauth.md
weppos: Fixed

PR #1043, opened 2026-06-23
Conform llms.txt to the llmstxt.org spec and list all official clients

aeden:
## What

Restructures the generated `llms.txt` (built from `content/llms.txt.erb`) so it conforms to the format grammar defined at <https://llmstxt.org/#format>, and expands the client list to cover every official DNSimple client plus the CLI.

## Why

The previous file was readable but off-spec: a strict llms.txt parser segments on H2 headers and expects each H2 section to be a link-only "file list", with free prose confined to the content block between the blockquote and the first H2. The original used horizontal rules as separators, a nested H3, prose-only and metadata H2 sections, all of which break that grammar. The Client Libraries section also only listed three of the nine official clients and omitted the CLI.

### Spec conformance
- **Remove `---` horizontal rules** — H2 headers are the only valid section delimiters.
- **Move `## About`, the `_Last updated/Contact/Website_` metadata, and `## Recommended Use for LLMs`** into the content block after the blockquote, where free prose belongs.
- **Promote `### Core Endpoints` to its own `## Core Endpoints`** — headings aren't allowed inside a section's content.
- **Rename `## Developer Policies` → `## Optional`** — the spec's reserved section for links that can be skipped for a shorter context (terms, rate limits, deprecations).
- **Add brief `: notes`** to reference links (optional in the format, helpful for LLM consumers).

### Content
- **List all nine official API v2 clients** (Ruby, Go, Elixir, Node.js, Java, C#, PHP, Python, Rust) using the canonical `dnsimple.link/api-client-*` redirects from `content/libraries.md`, plus a link to the **DNSimple CLI**. Previously only Ruby, Node.js, and Python were listed.
- **Refresh the `Last updated` date** to June 23, 2026.

All original links and the H1 + blockquote are preserved.

## Note on the "last updated" date

The date is kept as a fixed string in the template (updated manually with content changes) rather than rendered dynamically from the build/git date.

### DIFF
--- a/content/llms.txt.erb
+++ b/content/llms.txt.erb
@@ -2,7 +2,7 @@
 
 > The DNSimple Developer Portal provides reference documentation for the DNSimple API, webhooks, and integrations used to automate domain and DNS management.
 
-The portal documents the REST API, client libraries, webhooks, and automation workflows that power programmatic access to domains, DNS zones, and certificates. Last updated October 31, 2025. Contact support@dnsimple.com or visit https://developer.dnsimple.com.
+The portal documents the REST API, client libraries, webhooks, and automation workflows that power programmatic access to domains, DNS zones, and certificates. Last updated June 23, 2026. Contact support@dnsimple.com or visit https://developer.dnsimple.com.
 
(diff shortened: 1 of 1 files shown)


laetitiaeden (approved):
Thank you @aeden

gronk-droid (approved):

PR #1068, opened 2026-08-03
Document the domain.registration_notice webhook event

juankuquintana:
Belongs to https://github.com/dnsimple/dnsimple-business/issues/2863

Documents the new `domain.registration_notice` webhook event added in dnsimple/dnsimple-app#36353. The event fires when a registry sends an informational notice about a domain that requires action by the account; the initial notice families are the .DE holder-verification pair (`de_holder_verification_notice`, `de_holder_verification_suspension`).

## :mag: QA

- https://editor.swagger.io/ displays the new event schema correctly

## :clipboard: Deployment Pre/Post tasks

- [x] PRE: dnsimple/dnsimple-app#36353 and dnsimple/dnsimple-app#36356 are released

## :shipit: Deployment Verification

- [x] QAd


fixtures/v2/webhooks/domain.registration_notice/example.http
```diff
@@ -0,0 +1,9 @@
+POST / HTTP/1.1
+host: example.com
+accept-encoding: gzip
+content-type: application/json
+user-agent: DNSimple-Webhook-Notifier/8fd98a84516b72d5b863b27f93fcbaef61f06c03
+content-length: 907
```
dallasread: The body is 886 bytes, not 907. The sibling fixtures are all byte-accurate (`domain.register` 662/662, `domain.delete` 667/667), and integrators replay these files against real HTTP parsers, where a long `content-length` hangs the read.

```suggestion
content-length: 886
```

content/v2/openapi.yml
```diff
@@ -4837,6 +4837,28 @@ components:
...
+        domain:
+          $ref: '#/components/schemas/Domain'
+        registration_notice:
+          type: object
+          required: [type, message]
+          properties:
+            type:
+              type: string
+              description: |
+                Notice family. Values currently emitted:
```
dallasread: Should we mention that this list is subject to change?

```suggestion
                Notice family. New values may be added at any time; treat an unknown value as informational and fall back to displaying `message` verbatim. Values currently emitted:
```

dallasread (approved):

content/v2/openapi.yml
stephanoodle: Yes

PR #1076, opened 2026-09-01
Document the registrant verification notice types

juankuquintana:
Belongs to https://github.com/dnsimple/dnsimple-business/issues/2934.

Documents the notice families DNSimple is adding to the `domain.registration_notice` webhook event, alongside the .DE holder-verification pair that shipped in #1068. The payload shape is unchanged; only the `type` values grow, which is what the schema said integrators should expect.

- `se_registrant_verification_notice` — the .SE registry requires the registrant data to be verified by a deadline (dnsimple/dnsimple-app#36716).
- `icann_email_verification_notice`, `icann_email_verification_suspension`, `icann_email_verification_reactivation` — the ICANN registrant email verification workflow that covers every gTLD, from the notice that precedes suspension through the reactivation once the registrant verifies (dnsimple/dnsimple-app#36717).


dallasread (approved):
Thanks!

